PR #99, opened 2021-06-03
Move toolbelt/source:open to plugin-source

maggiben:
### What does this PR do?
Adds the command `force:org:open`

@W-9236226@


test/commands/source/open.test.ts
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { $$, expect, test } from '@salesforce/command/lib/test';
```
shetzel, line 7: I'd rather the source plugin not use this method of testing because it seems to have interactions with other stubs and spies.  If you look at the other command tests in the plugin there is a pattern to follow that I think works better.

shetzel (changes requested):
There are a large amount of changes to the yarn.lock.  Is that all just due to adding the open library or are there other unintended changes?

src/commands/force/source/open.ts
```diff
@@ -0,0 +1,160 @@
...
+    const openPath = typeName === 'FlexiPage' ? await this.handleSupportedTypes() : await this.handleUnsupportedTypes();
+
+    this.openResult = await this.open(openPath);
+  }
+
+  private getTypeNameDefinitionByFileName(fsPath: string): string | undefined {
+    if (fs.fileExistsSync(fsPath)) {
+      const metadataResolver = new MetadataResolver();
+      const components: SourceComponent[] = metadataResolver.getComponentsFromPath(fsPath);
+      return components[0].type.name;
```
WillieRuemmele, line 71: is there guaranteed to be only 1 result here?

src/commands/force/source/open.ts
```diff
@@ -0,0 +1,160 @@
...
+    type: string;
+    url: string;
+  };
+  Id: string;
+}
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-source', 'open');
+
+export class Open extends SfdxCommand {
```
WillieRuemmele, line 35: can we make this extend `SourceCommand`?

src/commands/force/source/open.ts
```diff
@@ -0,0 +1,160 @@
...
+    return url;
+  }
+
+  private resolveSuccess(): void {
+    if (!getString(this.openResult, 'url')) {
+      process.exitCode = 1;
+    }
+  }
+
+  private isJsonOutput(): boolean {
```
WillieRuemmele, line 103: a few of these methods will be built into `SourceCommand`

maggiben: Done

src/commands/force/source/open.ts
maggiben: yes we can.

src/commands/force/source/open.ts
maggiben: `getComponentsFromPath` always returns an array even for directories, we pick the first component type out of that list, ideally user should pick a single `flexipage`.
If file/directory exist but cannot determine source type the method throws an error `Could not infer a metadata type`

shetzel (approved):

PR #141, opened 2021-07-09
Sfdx user progress bar status bug

vamsimundra:
### What does this PR do?
Adds the missing functionality of displaying the progress of status when SFDX_USE_PROGRESS_BAR is set to false.
@W-9543714@

### DIFF
--- a/src/commands/force/source/deploy.ts
+++ b/src/commands/force/source/deploy.ts
@@ -4,15 +4,15 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
 import * as os from 'os';
+import * as chalk from 'chalk';
 import { flags, FlagsConfig } from '@salesforce/command';
 import { Messages } from '@salesforce/core';
 import { AsyncResult, DeployResult, MetadataApiDeploy } from '@salesforce/source-deploy-retrieve';
 import { Duration } from '@salesforce/kit';
 import { getString, isString } from '@salesforce/ts-types';
 import { env, once } from '@salesforce/kit';
-import { RequestStatus } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
+import { MetadataApiDeployStatus, RequestStatus } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
 import { DeployCommand } from '../../../deployCommand';
 import { ComponentSetBuilder } from '../../../componentSetBuilder';
 import { DeployResultFormatter, DeployCommandResult } from '../../../formatters/deployResultFormatter';
@@ -155,10 +155,10 @@ export class Deploy extends DeployCommand {
       this.updateDeployId(deploy.id);
 
       if (!this.isAsync) {
-        // if SFDX_USE_PROGRESS_BAR is unset or true (default true) AND we're not print JSON output
-        if (env.getBoolean('SFDX_USE_PROGRESS_BAR', true) && !this.isJsonOutput()) {
-          this.initProgressBar();
-          this.progress(deploy);
+        // we're not print JSON output
+        if (!this.isJsonOutput()) {
+          if (env.getBoolean('SFDX_USE_PROGRESS_BAR', true)) this.initProgressBar();
+          this.progress(deploy, env.getBoolean('SFDX_USE_PROGRESS_BAR', true));
         }
         this.deployResult = await deploy.pollStatus(500, waitDuration.seconds);
       }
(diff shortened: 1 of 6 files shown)


test/commands/source/deploy.test.ts
```diff
@@ -376,7 +381,8 @@ describe('force:source:deploy', () => {
         ensureCreateComponentSetArgs({ sourcepath });
         ensureDeployArgs();
         ensureHookArgs();
-        ensureProgressBar(0);
+        // this should not be ensurePrintStatus(0).
+        ensurePrintStatus(0);
```
vamsimundra, line 385: According to functionality, it should be ensurePrintStatus should not be 0. Because it should be getting called when  progress_bar is set false. I am trying to figure out why.
Am I missing anything here?

src/commands/force/source/deploy.ts
```diff
@@ -155,10 +155,10 @@ export class Deploy extends DeployCommand {
       this.updateDeployId(deploy.id);
 
       if (!this.isAsync) {
-        // if SFDX_USE_PROGRESS_BAR is unset or true (default true) AND we're not print JSON output
-        if (env.getBoolean('SFDX_USE_PROGRESS_BAR', true) && !this.isJsonOutput()) {
-          this.initProgressBar();
-          this.progress(deploy);
+        // we're not print JSON output
+        if (!this.isJsonOutput()) {
+          if (env.getBoolean('SFDX_USE_PROGRESS_BAR', true)) this.initProgressBar();
```
WillieRuemmele, line 160: can we move the `this.initProgressBar` call into `progress`? It would be nice to consolidate all of the relevant logic

src/commands/force/source/deploy.ts
```diff
@@ -225,37 +225,79 @@ export class Deploy extends DeployCommand {
...
+    } else {
+      if (data.completedDate) {
+        const deployStart: number = new Date(data.createdDate).getTime();
+        const deployEnd: number = new Date(data.completedDate).getTime();
+        this.ux.log(`Deployment finished in ${deployEnd - deployStart}ms`);
+      }
+      this.ux.log('');
+      const successHeader: string = chalk.green('Result');
+      const failureHeader: string = chalk.red('Result');
+      const header: string = data.success ? successHeader : failureHeader;
```
WillieRuemmele, line 242: Nit: could you we simplify this to `const header: string = data.success ? chalk.green('Result') : chalk.red('Result');`

src/commands/force/source/deploy.ts
```diff
@@ -225,37 +225,79 @@ export class Deploy extends DeployCommand {
...
+        }
       }
     });
-
     // any thing else should stop the progress bar
     deploy.onFinish((data) => {
-      // the final tick of `onUpdate` is actually fired with `onFinish`
-      this.progressBar.update(data.response.numberComponentsDeployed + data.response.numberTestsCompleted);
-      this.progressBar.stop();
+      if (progressBar) {
```
WillieRuemmele, line 285: can we change this so it's the same conditional statement between both listeners. e.g. always use `if(progressBar)`

src/commands/force/source/deploy.ts
```diff
@@ -225,37 +225,79 @@ export class Deploy extends DeployCommand {
...
+        this.progressBar.stop();
+      } else {
+        // Prints the final Status
+        this.printStatus(data.response);
+      }
     });
 
     deploy.onCancel(() => {
-      this.progressBar.stop();
+      if (progressBar) this.progressBar.stop();
```
WillieRuemmele, line 296: does toolbelt print anything for these events without the progress bar?

test/commands/source/deploy.test.ts
WillieRuemmele: it's getting called with `--json` which skips the printing step

src/commands/force/source/deploy.ts
vamsimundra: Nope. Based on what I found. They are not printing anything different.

shetzel:
@vamsimundra @WillieRuemmele 
I like the ProgressFormatter approach.  In fact, I think we can do more with it especially if retrieve also uses the progress bar.  Progress bar is a display element so most of the logic should go there.  Maybe we can have a ProgressFormatter base class, then a ProgressBarFormatter and a ProgressStatusFormatter.  Then the deploy command class would simply create the proper progress formatter based on the env var.

Also, I don't really like the toolbelt progress status output.  It's too verbose and gives redundant info.  We don't have to be backwards compatible with this output so I'd like to hear from Jochen about it.  I think we can sum up the relevant info in 2 lines; deploy data and test data.  The rest is not too useful.  See the `DeployReportResultFormatter` lines 33 and 34.  Or even on 1 line if we do something like on line 28 and provide a key for what the format means.

src/formatters/deployProgressBarFormatter.ts
```diff
@@ -0,0 +1,67 @@
...
+import cli from 'cli-ux';
+import { ProgressBar } from '../sourceCommand';
+import { ProgressFormatter } from './progressFormatter';
+export class DeployProgressBarFormatter extends ProgressFormatter {
+  protected progressBar?: ProgressBar;
+  public constructor(logger: Logger, ux: UX) {
+    super(logger, ux);
+  }
+
+  // displays the porgress of the Deployment
```
WillieRuemmele, line 20: NIT typo `porgress` `progress`

src/formatters/progressFormatter.ts
```diff
@@ -19,38 +18,5 @@ export class ProgressFormatter {
...
-    this.ux.log(`Component errors:  ${data.numberComponentErrors}`);
-    this.ux.log(successfulComponentsMessage);
-    this.ux.log(`Components total:  ${data.numberComponentsTotal}`);
-    this.ux.log(`Tests errors:  ${data.numberTestErrors}`);
-    this.ux.log(`Tests completed:  ${data.numberTestsCompleted}`);
-    this.ux.log(`Tests total:  ${data.numberTestsTotal}`);
-    this.ux.log(`Check only: ${data.checkOnly}`);
-    this.ux.log('');
-  }
+  public abstract progress(deploy: MetadataApiDeploy): void;
```
WillieRuemmele, line 21: exactly what I was hoping to see here

src/formatters/deployProgressStatusFormatter.ts
```diff
@@ -0,0 +1,69 @@
...
+      this.ux.log('');
+      const successHeader: string = chalk.green('Result');
+      const failureHeader: string = chalk.red('Result');
+      const header: string = data.success ? successHeader : failureHeader;
+      this.ux.styledHeader(header);
+      this.ux.log('');
+    }
+    const componentsTotal = getNumber(data, 'numberComponentsTotal');
+    if (componentsTotal) {
+      const componentsDeployed = getNumber(data, 'numberComponentsDeployed');
```
WillieRuemmele, line 53: what's the status of changing this info with Pooja and Juliet?

test/commands/source/deploy.test.ts
```diff
@@ -57,8 +58,10 @@ describe('force:source:deploy', () => {
   // Stubs
   let buildComponentSetStub: sinon.SinonStub;
   let initProgressBarStub: sinon.SinonStub;
-  let progressStub: sinon.SinonStub;
-  let progressWithNoBar: sinon.SinonStub;
+  let progressBarStub: sinon.SinonStub;
+  let progressStatusStub: sinon.SinonStub;
+  // let progressStub: sinon.SinonStub;
+  // let progressWithNoBar: sinon.SinonStub;
```
WillieRuemmele, line 64: let's remove these, and the other commented out assertion below ~ line 182

src/formatters/deployResultFormatter.ts
```diff
@@ -71,9 +71,11 @@ export class DeployResultFormatter extends ResultFormatter {
...
       throw new SfdxError(messages.getMessage('deployCanceled', [canceledByName]), 'DeployFailed');
     }
-    this.displaySuccesses();
-    this.displayFailures();
-    this.displayTestResults();
+    setTimeout(() => {
+      this.displaySuccesses();
+      this.displayFailures();
+      this.displayTestResults();
+    }, 500);
```
WillieRuemmele, line 78: you might need to make the `progress` method async and wrap everything inside of it in a Promise, and await that Promise

WillieRuemmele (approved):

src/formatters/deployProgressStatusFormatter.ts
```diff
@@ -0,0 +1,76 @@
...
+    deploy.onFinish((data) => {
+      this.printDeployStatus(data.response);
+    });
+    deploy.onError((error: Error) => {
+      throw error;
+    });
+  }
+
+  // Prints Deploying status
+  private printDeployStatus(data: MetadataApiDeployStatus): void {
```
shetzel, line 41: This is very similar to `DeployReportResultFormatter`.  I think there is some code sharing that can happen, or maybe this code can even be removed and it can just use that other formatter.  Less code means less to maintain, less chance for bugs, and more consistency in the output between commands.

Let's keep this as is but create a work item to handle possible refactoring.

shetzel (approved):
This is great!

PR #146, opened 2021-07-13
fix: deploy errors are reported properly

shetzel:
### What does this PR do?
Deploy errors are in the JSON output but not being displayed in stdout in all cases. This was due to the code assuming that the response contained `DeployMessage[]` but it can also (unfortunately) return `DeployMessage`. 

Deploy errors are now displayed in the stdout table format.

@W-9596635@


WillieRuemmele (approved):
nice new formatter tests

mshanemc (approved):

PR #149, opened 2021-07-19
fix: retrieve package names output

WillieRuemmele:
### What does this PR do?
fixes retrieves with package name's output to list the package name, path, and source retrieved into that dir
matches toolbelt with the package section and JSON
differs from toolbelt by including the `Retrieved Source` section

@W-9584557@

### DIFF
--- a/src/commands/force/source/retrieve.ts
+++ b/src/commands/force/source/retrieve.ts
@@ -6,14 +6,19 @@
  */
 
 import * as os from 'os';
+import { join } from 'path';
 import { flags, FlagsConfig } from '@salesforce/command';
-import { Messages } from '@salesforce/core';
+import { Messages, SfdxProject } from '@salesforce/core';
 import { Duration } from '@salesforce/kit';
 import { getString } from '@salesforce/ts-types';
 import { RetrieveResult } from '@salesforce/source-deploy-retrieve';
 import { RequestStatus } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
 import { SourceCommand } from '../../../sourceCommand';
-import { RetrieveResultFormatter, RetrieveCommandResult } from '../../../formatters/retrieveResultFormatter';
+import {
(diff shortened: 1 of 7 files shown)


src/commands/force/source/retrieve.ts
```diff
@@ -105,12 +110,20 @@ export class Retrieve extends SourceCommand {
...
+    this.getFlag<string[]>('packagenames', []).forEach((name) => {
+      packages.push({ name, path: join(projectPath, name) });
+    });
+
     const formatterOptions = {
       waitTime: this.getFlag<Duration>('wait').quantity,
       verbose: this.getFlag<boolean>('verbose', false),
     };
     const formatter = new RetrieveResultFormatter(this.logger, this.ux, formatterOptions, this.retrieveResult);
+    formatter.packages = packages;
```
shetzel, line 126: What do you think about changing the `formatterOptions` type to add an optional `packages` option instead?

shetzel (commented):
there are a ton of changes in the yarn.lock file.  Maybe we can leave it alone for this PR and handle those updates when we do all the dependabot merges?

src/commands/force/source/retrieve.ts
WillieRuemmele: I was doing that at first, but that type is defined in the `ResultFormatter` class, which is a base class for `deploy` and `retrieve`, but `retrieve` only has packagenames. It felt awkward altering the base class for only one of the child classes... but I'm open to doing that way too

shetzel: Why can't we do this:
```javascript
export interface RetrieveResultFormatterOptions extends ResultFormatterOptions {
  packages?: string;
}
```

and then:
`public constructor(logger: Logger, ux: UX, options: RetrieveResultFormatterOptions, result: RetrieveResult) {`

shetzel (approved):

PR #151, opened 2021-07-22
fix: add verbose test detail when deploying with tests

WillieRuemmele:
### What does this PR do?
adds verbose testing output on deploy
@W-9622493@

requires https://github.com/forcedotcom/source-deploy-retrieve/pull/395


mshanemc (approved):

src/formatters/deployResultFormatter.ts
```diff
@@ -8,13 +8,17 @@
...
 import { DeployResult } from '@salesforce/source-deploy-retrieve';
 import {
+  CodeCoverage,
   FileResponse,
+  LocationsNotCovered,
   MetadataApiDeployStatus,
   RequestStatus,
+  Successes,
 } from '@salesforce/source-deploy-retrieve/lib/src/client/types';
+import { normalizeToArray } from '@salesforce/source-deploy-retrieve/lib/src/utils';
```
shetzel, line 21: This is a bad practice.  If it's not exported at the top level then we should not rely on it.  They could rename it or move it and it would break us but that wouldn't be considered a breaking change to the library because it's not part of their public API.  Let's instead just have our own, or add it to the kit library and use it.

src/formatters/deployResultFormatter.ts
```diff
@@ -155,4 +161,100 @@ export class DeployResultFormatter extends ResultFormatter {
...
+              }
+
+              return color(`${pctCovered}%`);
+            },
+          },
+          {
+            key: 'locationsNotCovered',
+            label: 'Uncovered Lines',
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore the format method expects a string, we're passing in LocationsNotCovered[]
```
shetzel, line 236: What if we build the string separately and set it on `format` so we don't need the `@ts-ignore`

src/formatters/deployResultFormatter.ts
```diff
@@ -155,4 +161,100 @@ export class DeployResultFormatter extends ResultFormatter {
...
+            format: (locationsNotCovered: LocationsNotCovered[]): string => {
+              if (!locationsNotCovered) {
+                return '';
+              }
+              // asArray wasn't properly converting a single LocationsNotCovered object to [LocationsNotCovered]
+              const locations = Array.isArray(locationsNotCovered) ? locationsNotCovered : [locationsNotCovered];
```
shetzel, line 242: Use the `normalizeToArray` function and remove the comment.

src/formatters/resultFormatter.ts
```diff
@@ -50,6 +51,15 @@ export abstract class ResultFormatter {
     });
   }
 
+  protected sortTestResults(results: Failures[] | Successes[] = []): Failures[] | Successes[] {
+    return results.sort((a: Successes, b: Successes) => {
```
shetzel, line 55: The typing here is a bit wonky and I'm surprised TS doesn't complain.  Assuming SDR doesn't provide a common base type for `Failure` and `Success`, it would be better to have 1 type here that only used `name`, and `methodName` properties.

shetzel (changes requested):

shetzel (approved):

PR #155, opened 2021-07-26
fix: missing metadata type error matches toolbelt

WillieRuemmele:
### What does this PR do?
catches and transforms the SDR error for a missing metadata type to match the toolbelt's

should hopefully defer blame to the user rather than us
@W-9657211@

### DIFF
--- a/src/componentSetBuilder.ts
+++ b/src/componentSetBuilder.ts
@@ -41,61 +41,71 @@ export class ComponentSetBuilder {
     const csAggregator: ComponentLike[] = [];
 
     const { sourcepath, manifest, metadata, packagenames, apiversion, sourceapiversion } = options;
-
-    if (sourcepath) {
-      logger.debug(`Building ComponentSet from sourcepath: ${sourcepath.toString()}`);
-      sourcepath.forEach((filepath) => {
-        if (fs.fileExistsSync(filepath)) {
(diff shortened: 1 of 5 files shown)


src/componentSetBuilder.ts
```diff
@@ -41,61 +41,71 @@ export class ComponentSetBuilder {
...
-      csAggregator.push(...finalized);
+        const directoryPaths = options.metadata.directoryPaths;
+        logger.debug(`Searching for matching metadata in directories: ${directoryPaths.join(', ')}`);
+        const fromSource = ComponentSet.fromSource({ fsPaths: directoryPaths, include: filter });
+        // If no matching metadata is found, default to the original component set
+        const finalized = fromSource.size > 0 ? fromSource : filter;
+        csAggregator.push(...finalized);
+      }
+    } catch (e) {
+      if ((e as Error).message.includes('Missing metadata type definition in registry for id')) {
```
mshanemc, line 101: Are we logging this via telemetry (it could be that the regular exception logging picks them up ok)

If not, it'd be great to have these as a type of exception, with the mdtype(s) in there, so we can use that data to help MD owners know customers are trying to deploy their feature AND help us prioritize support gaps.

mshanemc (approved):
ok to defer for now, but if you do, create a nice WI that points to this section of code so we remember where it should go

PR #168, opened 2021-08-03
Wr/manifest generate

WillieRuemmele:
### What does this PR do?
adds the `sfdx force:source:manifest:create` command.

I think that the NUT coverage should be enough, the command is so small that unit tests seemed extra when the `ComponentSetBuilder` is already covered
@W-9608078@


src/commands/force/source/manifest/create.ts
```diff
@@ -0,0 +1,115 @@
...
+      description: messages.getMessage('flags.manifesttype'),
+      options: Object.keys(manifestTypes),
+      char: 't',
+    }),
+    outputdir: flags.string({
+      char: 'o',
+      description: messages.getMessage('flags.outputdir'),
+    }),
+  };
+  public requiredFlags = ['metadata', 'sourcepath'];
```
shetzel, line 60: I think a better name for this would be xorFlags, meaning 1 and only 1 of them must be set when running the command.  I think this should really move into oclif at some point too.

src/sourceCommand.ts
```diff
@@ -21,6 +21,7 @@ export type ProgressBar = {
 
 export abstract class SourceCommand extends SfdxCommand {
   public static readonly DEFAULT_SRC_WAIT_MINUTES = 33;
+  public requiredFlags: string[] = [];
```
shetzel, line 24: I think this should be protected not public.

src/sourceCommand.ts
```diff
@@ -30,6 +31,15 @@ export abstract class SourceCommand extends SfdxCommand {
     return getBoolean(this.flags, 'json', false);
   }
 
+  protected validateFlags(flags: string[]): void {
```
shetzel, line 34: We don't need to pass in the flags.  SourceCommand has them as `this.flags`

test/nuts/create.nut.ts
```diff
@@ -0,0 +1,108 @@
...
+  let session: TestSession;
+
+  before(async () => {
+    session = await TestSession.create({
+      project: {
+        gitClone: 'https://github.com/trailheadapps/dreamhouse-lwc.git',
+      },
+      setupCommands: [
+        'sfdx force:org:create -f config/project-scratch-def.json --setdefaultusername --wait 10 --durationdays 1',
+        'sfdx force:source:deploy -p force-app',
```
shetzel, line 41: Why does the test need to create an org and deploy as part of setup?

test/nuts/create.nut.ts
```diff
@@ -0,0 +1,108 @@
...
+  '        <members>GeocodingService</members>\n' +
+  '        <members>GeocodingServiceTest</members>\n' +
+  '        <members>PagedResult</members>\n' +
+  '        <members>PropertyController</members>\n' +
+  '        <members>SampleDataController</members>\n' +
+  '        <members>TestPropertyController</members>\n' +
+  '        <members>TestSampleDataController</members>\n' +
+  '        <name>ApexClass</name>\n' +
+  '    </types>\n' +
+  '    <version>52.0</version>\n' +
```
shetzel, line 28: Unless you override the API version this will fail as soon as 52.0 is not current.

shetzel (changes requested):

test/nuts/create.nut.ts
WillieRuemmele: I figured that way it would validate that the generated manifest was would work with a deploy. SDR should be verifying that `getPackageXml()` returns a valid package... but I thought it would be pretty easy to do here too

command-snapshot.json
```diff
@@ -34,7 +34,22 @@
...
+    "flags": [
+      "apiversion",
+      "json",
+      "loglevel",
+      "manifestname",
+      "manifesttype",
+      "metadata",
+      "outputdir",
+      "sourcepath",
+      "targetusername"
```
WillieRuemmele, line 51: @shetzel to use `SfdxCommand`'s standardized `--apiversion` flag I had to set `supportsUsername=true` which also includes the `--targetusername`... Do you think it would be better to duplicate the `apiversion` flag?

shetzel: Unless the command will use that flag (-u) I would not set `supportsUsername=true`.  Just duplicate the apiversion flag.  There might be a way to reference it without duplicating.

shetzel (approved):
Nice!

PR #173, opened 2021-08-10
Wr/convert package names

WillieRuemmele:
### What does this PR do?
`source:convert` now matches toolbelt's default behavior with default `--outputdir` and a specified `--outputdir`
@W-9735120@

### DIFF
--- a/src/commands/force/source/convert.ts
+++ b/src/commands/force/source/convert.ts
@@ -95,7 +95,8 @@ export class Convert extends SourceCommand {
     });
 
     const packageName = this.getFlag<string>('packagename');
-    const outputDirectory = this.getFlag<string>('outputdir');
+    const outputDirectory =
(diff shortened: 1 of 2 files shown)


src/commands/force/source/convert.ts
```diff
@@ -93,13 +94,29 @@ export class Convert extends SourceCommand {
       },
     });
 
+    const packageName = this.getFlag<string>('packagename');
+    const outputDirectory =
+      this.getFlag<string>('outputdir') === './' ? `metadataPackage_${Date.now()}` : this.getFlag<string>('outputdir');
```
shetzel, line 99: Maybe we should resolve the outputdir.  A value of `'.'` here throws an error.  We might also want to ensure any directory specified here is empty before continuing or an error will be thrown during rename.

shetzel (changes requested):
A few small changes requested.

PR #176, opened 2021-08-16
chore: code refactoring of exports

vamsimundra:
### What does this PR do?
Made changes in how we export types from SDR.
@W-9727845@

### DIFF
--- a/src/componentSetBuilder.ts
+++ b/src/componentSetBuilder.ts
@@ -7,7 +7,7 @@
 
 import * as path from 'path';
 import { ComponentSet, RegistryAccess } from '@salesforce/source-deploy-retrieve';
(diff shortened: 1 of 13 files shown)


src/componentSetBuilder.ts
```diff
@@ -7,7 +7,7 @@
 
 import * as path from 'path';
 import { ComponentSet, RegistryAccess } from '@salesforce/source-deploy-retrieve';
-import { ComponentLike } from '@salesforce/source-deploy-retrieve/lib/src/resolve/types';
+import { ComponentLike } from '@salesforce/source-deploy-retrieve';
```
WillieRuemmele, line 10: line 9 and 10 should be combined, this should be applied to every duplicate import

WillieRuemmele (approved):

PR #179, opened 2021-08-24
fix: trim metadata type, and name to prevent unnecessary errors

WillieRuemmele:
### What does this PR do?
`deploy -m "ApexClass: GeocodingService" would cause a `no source backed components` Error, trim the metadata type and name to ensure we're parsing everything correctly

@W-9803972@  https://github.com/forcedotcom/cli/issues/1151

### DIFF
--- a/src/componentSetBuilder.ts
+++ b/src/componentSetBuilder.ts
@@ -82,10 +82,10 @@ export class ComponentSetBuilder {
         metadata.metadataEntries.forEach((rawEntry) => {
           const splitEntry = rawEntry.split(':');
           // The registry will throw if it doesn't know what this type is.
-          registry.getTypeByName(splitEntry[0]);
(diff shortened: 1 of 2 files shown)


src/componentSetBuilder.ts
```diff
@@ -82,10 +82,10 @@ export class ComponentSetBuilder {
         metadata.metadataEntries.forEach((rawEntry) => {
           const splitEntry = rawEntry.split(':');
```
mshanemc, line 83:
```suggestion
         const splitEntry = rawEntry.split(':').map(entry => entry.trim())
```

will trim both, so you don't have to trim in 2 places OR worry about the untrimmed type on line 87

mshanemc (approved):
approved, with suggestion

shetzel (approved):

PR #180, opened 2021-08-30
Wr/nuts check deploy

WillieRuemmele:
### What does this PR do?
verifies "deployed" files in the NUTs are actually deployed by parsing the Deploy Results
@W-9812177@ waiting on https://github.com/salesforcecli/source-testkit/pull/8


test/nuts/seeds/mpd.deploy.seed.ts
```diff
@@ -38,31 +38,31 @@ context('MPD Deploy NUTs [exec: %EXECUTABLE%]', () => {
     describe('--sourcepath', () => {
       it('should deploy all CustomLabels from a single package', async () => {
         await testkit.deploy({ args: `--sourcepath ${path.join('force-app', 'main', 'default', 'labels')}` });
-        await testkit.expect.filesToBeChanged([forceAppLabels]);
+        await testkit.expect.filesToBeDeployed([forceAppLabels]);
```
mshanemc, line 41: so the original `filesToBeDeployed` is ok for most everything else?

mshanemc (approved):
ready to see this go green

test/nuts/seeds/mpd.deploy.seed.ts
WillieRuemmele: yes, it was only an issue when you would deploy `force-app` and then a subset of `force-app` because the subset deploy wouldn't increment the  `RevisionCounter`

PR #199, opened 2021-09-01
Wr/delete

WillieRuemmele:
### What does this PR do?
adds the `source:delete` command
adds unit tests
adds `delete` NUTs

@W-9137776@
@W-7552573@


src/commands/force/source/delete.ts
```diff
@@ -0,0 +1,206 @@
...
+  protected formatResult(): DeployCommandResult {
+    const formatterOptions = {
+      verbose: this.getFlag<boolean>('verbose', false),
+    };
+
+    const formatter = new DeleteResultFormatter(this.logger, this.ux, formatterOptions, this.deployResult);
+
+    // The DeleteResultFormatter will use SDR and scan the directory, if the files have been deleted, it will throw an error
+    // so we'll delete the files locally now
+    this.deleteFilesLocally();
```
shetzel, line 167: This feels out of place in a results formatting function, and I don't think the comment applies anymore.  What if we move this to either the `resolveSuccess` method or the `run` method  just after calling `resolveSuccess()`?

src/commands/force/source/delete.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    this.sourceComponents = this.componentSet.getSourceComponents().toArray();
+
+    if (!this.sourceComponents.length) {
+      // if we didn't find any components to delete, let the user know and exit
+      // matches toolbelt
+      this.ux.styledHeader(chalk.blue('Deleted Source'));
+      this.ux.log('No results found');
+      this.exit(0);
+    }
```
shetzel, line 108: I think we should keep formatting to the formatters unless it's not possible.  Also, we should avoid calling exit explicitly and let the command framework handle it.  What if we just did `return;` here and let the formatter display this?

src/commands/force/source/delete.ts
```diff
@@ -0,0 +1,206 @@
...
+    const cs = new ComponentSet([]);
+    this.sourceComponents.map((component) => {
+      cs.add(component, true);
+    });
+    this.componentSet = cs;
+
+    if (!(await this.handlePrompt())) {
+      // the user said 'no' to the prompt
+      this.exit(0);
+    }
```
shetzel, line 120: What if we just used a private for this and `resolveSuccess` would check that too?  Maybe: `this.canceled = true; return;`

src/commands/force/source/delete.ts
```diff
@@ -0,0 +1,206 @@
...
+      this.ux.log('No results found');
+      this.exit(0);
+    }
+
+    // create a new ComponentSet and mark everything for deletion
+    const cs = new ComponentSet([]);
+    this.sourceComponents.map((component) => {
+      cs.add(component, true);
+    });
+    this.componentSet = cs;
```
shetzel, line 115: We can't do it this way because we'll lose `apiVersion` and `sourceApiVersion` by creating a new `ComponentSet`.  We should modify `ComponentSetBuilder` to take another optional option to mark them for delete.

shetzel (changes requested):

src/commands/force/source/delete.ts
WillieRuemmele: [destructive change files](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_deploy_deleting_files.htm) don't include an api version in them, it's recommend to use an empty package, with the api, if necessary

shetzel: Ok but I still think creating multiple ComponentSets is an anti-pattern.  If we don't think other commands will need ComponentSetBuilder modified for deletes then this is ok.

src/commands/force/source/delete.ts
```diff
@@ -0,0 +1,204 @@
...
+        }
+      });
+    }
+  }
+
+  private async handlePrompt(): Promise<boolean> {
+    if (!this.getFlag('noprompt')) {
+      const paths = this.sourceComponents.map((component) => component.content + '\n' + component.xml).join('\n');
+      const promptMessage = messages.getMessage('prompt', [paths]);
+      const answer = (await prompt(promptMessage)) as string;
```
mshanemc, line 199: I've always wondered why CLI commands use `prompt` instead of `confirm` (and then have to do the extra casing, etc).

source: https://github.com/oclif/cli-ux/blob/fd0ca55f2ae7096ea273592be35e071ee4ddcb0d/src/prompt.ts#L132

ex in use
https://github.com/forcedotcom/source-tracking/blob/e73b17609ae410be4043b423b8488968a7513afe/src/commands/force/source/tracking/reset.ts#L44

WillieRuemmele: Oh, it's probably just a carry-over from TB, I'll swap it for `confirm`

shetzel (approved):

mshanemc:
QA Notes:
I'm using this repo: https://github.com/SFDC-Assets/flow-builder

after running its setup (orgInit.sh)

I'm getting some undefined in the warnings
```
➜  flow-builder git:(master) ✗ sfdx force:source:delete -c -m CustomTab
This operation will delete the following files on your computer and in your org: 
undefined
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/tabs/Assigned_Training__c.tab-meta.xml
undefined
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/tabs/Employee__c.tab-meta.xml
undefined
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/tabs/Equipment__c.tab-meta.xml
```

mshanemc:
QA Notes: same `undefined` with a specifed file 
```
➜  flow-builder git:(master) ✗ sfdx force:source:delete -c -m CustomTab:Employee__c
This operation will delete the following files on your computer and in your org: 
undefined
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/tabs/Employee__c.tab-meta.xml
```

mshanemc:
QA notes:
`-c` yield the following stdout message...it's technically a deployment but that's a bit "weird" for a delete operation.  `Successfully validated the deployment. 1 components deployed and 0 tests run.`.

mshanemc:
QA Notes:
good deletes with -p, -m (tried by types, my specific files, even -m CustomField:Object.Field).
it's also deleting local source (as expected)

mshanemc:
QA Notes:
I deleted a file, then undeleted it on local FS, then ran delete on it again.  I get this non-useful error
```
ERROR running force:source:delete:  Deploy failed.
``` 

when what I really want is in the --json
```json
"componentFailures": {
        "changed": "false",
        "componentType": "CustomField",
        "created": "false",
        "createdDate": "2021-09-17T15:17:18.000Z",
        "deleted": "false",
        "fileName": "destructiveChanges.xml",
        "fullName": "destructiveChanges.xml",
        "problem": "No CustomField named: Employee__c.Onboarding_Buddy__c found",
        "problemType": "Warning",
        "success": "false"
      },
```

mshanemc:
QA Notes:
for decomposed source, it's presenting a somewhat misleading list of what's going to be deleted (relying on the user to know that the fields all roll up onto the object).  I think I would have expected a list of fields?

```
➜  flow-builder git:(master) ✗ sfdx force:source:delete -m CustomField:Employee__c.Onboarding_Buddy__c,CustomObject:Employee__c       
This operation will delete the following files on your computer and in your org: 
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/Employee__c.object-meta.xml
```

`toolbelt` gives the full list of deletions for the same command

```
➜  flow-builder git:(master) ✗ sfdx force:source:delete -m CustomField:Employee__c.Onboarding_Buddy__c,CustomObject:Employee__c
This operation will delete the following files on your computer and in your org: 
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/Employee__c.object-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/City__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Department__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Email__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/First_Name__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Last_Name__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Onboarding_Buddy__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Phone__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/PostalCode__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Role__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Send_Notification__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Start_Date__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/State__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Status__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/Street_Address__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/fields/User_Record__c.field-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/compactLayouts/Default.compactLayout-meta.xml
Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/listViews/All.listView-meta.xml
```

mshanemc:
QA Notes:
When adding a non-existent type/item and a pair that does exist, we get a warning that only refers to the existing one (but doesn't inform the user that the non-existent one is a mistake)

```
➜  flow-builder git:(master) ✗ sfdx force:source:delete -m CustomTab:Employee__x,CustomObject:Employee__c  
This operation will delete the following files on your computer and in your org: 
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c
/Users/shane.mclaughlin/eng/repros/flow-builder/force-app/main/default/objects/Employee__c/Employee__c.object-meta.xml
```

after confirming, it doesn't look like any attempt was made to delete the non-existent one (which seems fair).  I think this should have thrown an error before trying to delete.

[only applies to `-m` because `-p` checks for dir/file existing.]

mshanemc (approved):

WillieRuemmele:
https://github.com/forcedotcom/cli/issues/971

PR #209, opened 2021-09-20
Sm/source-tracking-beta

mshanemc:
### What does this PR do?
source tracking commands using STL

@W-9915867@


src/commands/force/source/beta/status.ts
```diff
@@ -0,0 +1,181 @@
...
+        outputRows = outputRows.map((row) =>
+          conflictKeys.includes(getKeyFromStrings(row.type, row.fullName))
+            ? { ...row, state: `${row.state} (Conflict)` }
+            : row
+        );
+      }
+    }
+
+    // sort order is state, type, fullname
+    outputRows.sort((a, b) => {
```
WillieRuemmele, line 129: should this sort of output stuff be moved into a formatter?

test/nuts/ebikes-lwc/.eslintignore
```diff
@@ -0,0 +1,13 @@
+**/lwc/**/*.css
```
WillieRuemmele, line 1: why do we need a clone of ebikes in the repo? Why can't we clone the project for NUTs?

messages/source_tracking.js
```diff
@@ -0,0 +1,24 @@
+const warning =
```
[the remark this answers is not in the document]
mshanemc: I saw the use of template literals in there and said, "oh, json can't reference itself so I'll leave it"

Could probably duplicate the warning, or import that as a message to pass to the other messages or something.

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,82 @@
...
+      projectPath: this.project.getPath(),
+      toValidate: 'plugin-source',
+      command: replaceRenamedCommands('force:source:pull'),
+    });
+
+    this.ux.startSpinner('Loading source tracking information');
+    const tracking = await SourceTracking.create({
+      org: this.org,
+      project: this.project,
+      apiVersion: this.flags.apiversion as string,
```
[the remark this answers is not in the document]
mshanemc: this.flags.apiversion can be undefined...so it'll try to get stuff from project and config files (and ultimately the org).  This is needed to support the flag.  I should add some comments about that.

🤪 if you used vscode you could review PRs from the editor

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,82 @@
...
+      this.ux.table(retrieveResult, {
+        columns: [
+          { label: 'STATE', key: 'state' },
+          { label: 'FULL NAME', key: 'fullName' },
+          { label: 'TYPE', key: 'type' },
+          { label: 'PROJECT PATH', key: 'filePath' },
+        ],
+      });
+    }
+    return retrieveResult.map(({ state, fullName, type, filePath }) => ({ state, fullName, type, filePath }));
```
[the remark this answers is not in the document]
mshanemc: the retrieveResult has more fields present, and I only wanted those three.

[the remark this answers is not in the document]
mshanemc: that's the destructured equivalent to 
```ts
result.map(item=>({ state: item.state, fullName: item.fullName, type: item.type, filePath: item.filePath}))

File: src/commands/force/source/beta/pull.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (WillieRuemmele) [reply src/commands/force/source/beta/pull.ts:56]: I'll start doing my reviews from WebStorm 😜

File: src/commands/force/source/beta/status.ts (hunk 1 above)
Comment (mshanemc) [reply src/commands/force/source/beta/status.ts:129]: that would be more consistent with the rest of plugin-source.

File: test/nuts/ebikes-lwc/.eslintignore (hunk 2 above)
Comment (mshanemc) [reply test/nuts/ebikes-lwc/.eslintignore:1]: I expect the sample apps to change over time.  Tracking nuts are (perhaps overly-) specific about how many rows there should be in various scenarios.

File: src/commands/force/source/beta/push.ts [hunk 6]
```diff
@@ -0,0 +1,148 @@
...
+    // there might have been components in local tracking, but they might be ignored by SDR or unresolvable.
+    // SDR will throw when you try to resolve them, so don't
+    if (componentSet.size === 0) {
+      this.logger.warn('There are no changes to deploy');
+      return;
+    }
+
+    // fire predeploy event for sync and async deploys
+    await this.lifecycle.emit('predeploy', componentSet.toArray());
+    this.ux.log(`*** Deploying with ${this.isRest ? 'REST' : 'SOAP'} API ***`);
```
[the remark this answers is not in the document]
mshanemc: good call

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+  ComponentStatus,
+} from '@salesforce/source-deploy-retrieve';
+import { SourceTracking, throwIfInvalid, replaceRenamedCommands, ChangeResult } from '@salesforce/source-tracking';
+import { processConflicts } from '../../../../formatters/conflicts';
+import { SourceCommand } from '../../../../sourceCommand';
+import { PullResponse, PullResultFormatter } from '../../../../formatters/pullFormatter';
+Messages.importMessagesDirectory(__dirname);
+const messages: Messages = Messages.loadMessages('@salesforce/plugin-source', 'pull');
+
+export default class SourcePull extends SourceCommand {
```
shetzel, line 26: After beta will this change to be named just `Pull`?

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+      min: Duration.minutes(0), // wait=0 means deploy is asynchronous
+      description: messages.getMessage('flags.waitLong'),
+    }),
+  };
+
+  protected static requiresUsername = true;
+  protected static requiresProject = true;
+  protected hidden = true;
+  protected readonly lifecycleEventNames = ['preretrieve', 'postretrieve'];
+  protected tracking: SourceTracking;
```
shetzel, line 47: nit: I think this should be `tracking!: SourceTracking;` since it's not set here or in the constructor.

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.tracking = await SourceTracking.create({
+      org: this.org,
+      project: this.project,
+    });
+
+    await this.tracking.ensureRemoteTracking(true);
+    this.ux.setSpinnerStatus('Checking for conflicts');
+
+    if (!this.flags.forceoverwrite) {
+      processConflicts(await this.tracking.getConflicts(), this.ux, messages.getMessage('sourceConflictDetected'));
```
shetzel, line 81: I don't love passing `this.ux` to STL.  I would expect STL to just return the conflicts and the command would handle it how it wants.  If it's this way so that code can be shared and the sf plugin won't need to also implement it, what will happen if sf (or any other plugin) wants to process conflicts differently, or even have a different message?

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+      projectPath: this.project.getPath(),
+      toValidate: 'plugin-source',
+      command: replaceRenamedCommands('force:source:pull'),
+    });
+
+    this.ux.startSpinner('Loading source tracking information');
+    this.tracking = await SourceTracking.create({
+      org: this.org,
+      project: this.project,
+    });
```
shetzel, line 75: I love this!

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  protected async preChecks(): Promise<void> {
+    // checks the source tracking file version and throws if they're toolbelt's old version
+    throwIfInvalid({
+      org: this.org,
+      projectPath: this.project.getPath(),
+      toValidate: 'plugin-source',
+      command: replaceRenamedCommands('force:source:pull'),
+    });
```
shetzel, line 69: It's a bit strange that STL does this but I understand the reasons why.  I think we should mark it in STL as deprecated from the start so we can eventually remove it after several releases.

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+  protected retrieveResult: RetrieveResult;
+  protected deleteFileResponses: FileResponse[];
+
+  public async run(): Promise<PullResponse[]> {
+    await this.preChecks();
+    await this.retrieve();
+    // do not parallelize delete and retrieve...we only get to delete IF retrieve was successful
+    await this.doDeletes(); // deletes includes its tracking file operations
+    await this.updateTrackingFilesWithRetrieve();
+    this.ux.stopSpinner();
```
shetzel, line 57: Is it ok to not stop the spinner in a try/finally?  Does it stop if the above code throws?

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  protected async doDeletes(): Promise<void> {
+    this.ux.setSpinnerStatus('Checking for deletes from the org and updating source tracking files');
+    const changesToDelete = await this.tracking.getChanges<SourceComponent>({
+      origin: 'remote',
+      state: 'delete',
+      format: 'SourceComponent',
+    });
+    this.deleteFileResponses = await this.tracking.deleteFilesAndUpdateTracking(changesToDelete);
```
shetzel, line 92: Is there a method to `deleteFilesAndNotUpdateTracking`?  Since you're calling a method on `this.tracking` I would assume that every action affects or updates tracking.

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  protected async updateTrackingFilesWithRetrieve(): Promise<void> {
+    this.ux.setSpinnerStatus('Updating source tracking files');
```
shetzel, line 96: I love all the UI updates via the spinner!

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.ux.setSpinnerStatus('Checking for deletes from the org and updating source tracking files');
+    const changesToDelete = await this.tracking.getChanges<SourceComponent>({
+      origin: 'remote',
+      state: 'delete',
+      format: 'SourceComponent',
+    });
+    this.deleteFileResponses = await this.tracking.deleteFilesAndUpdateTracking(changesToDelete);
+  }
+
+  protected async updateTrackingFilesWithRetrieve(): Promise<void> {
```
shetzel, line 95: Should we just call this `updateTracking` since we can assume it's from the retrieve result?

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+    await Promise.all([
+      // commit the local file successes that the retrieve modified
+      this.tracking.updateLocalTracking({
+        files: successes.map((fileResponse) => fileResponse.filePath).filter(Boolean),
+      }),
+      this.tracking.updateRemoteTracking(
+        successes.map(({ state, fullName, type, filePath }) => ({ state, fullName, type, filePath })),
+        true // skip polling because it's a pull
+      ),
+    ]);
```
shetzel, line 111: This is so awesome!

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+      output: this.project.getDefaultPackage().path,
+    });
+
+    this.ux.setSpinnerStatus('Retrieving metadata from the org');
+
+    // assume: remote deletes that get deleted locally don't fire hooks?
+    await this.lifecycle.emit('preretrieve', componentSet.toArray());
+    this.retrieveResult = await mdapiRetrieve.pollStatus(1000, this.getFlag<Duration>('wait').seconds);
+
+    // assume: remote deletes that get deleted locally don't fire hooks?
```
shetzel, line 151: > assume: remote deletes that get deleted locally don't fire hooks?
Good question.  Not sure what currently happens in source:pull but match that.  If it's not possible in the legacy version then we can do whatever makes sense.

shetzel (commented):
Only reviewed pull.  I'll continue asap

src/commands/force/source/beta/pull.ts
mshanemc: `processConflicts` is in the plugin's /formatters, not in STL.  It's passing output from `STL.getConflicts()` to the processConflicts "formatter" (and I use that term pretty loosely)

src/commands/force/source/beta/pull.ts
mshanemc: good idea.  stl has a Compatibility class for handling all transitional stuff (messaging, checking that you're not mixing files, etc).

src/commands/force/source/beta/pull.ts
mshanemc: I really wanted to do those on `retrieve` but didn't (focus!)

src/commands/force/source/beta/pull.ts
mshanemc: I did this because there's also the update of tracking files from the delete above (and you cold be doing either or both)

I'm open to change if my responses about those descriptive names isn't persuasive.

src/commands/force/source/beta/pull.ts
shetzel: Ahh, ok.  I missed that.

src/commands/force/source/beta/pull.ts
mshanemc: done here: https://github.com/forcedotcom/source-tracking/pull/20/commits/b68daec95b4fd9090e68eb28d15c96205a2b7dde

src/commands/force/source/beta/status.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export default class SourceStatus extends SfdxCommand {
+  public static description = messages.getMessage('description');
+  public static readonly examples = replaceRenamedCommands(messages.getMessage('examples')).split(os.EOL);
+  protected static flagsConfig: FlagsConfig = {
+    all: flags.boolean({
+      char: 'a',
+      description: messages.getMessage('flags.all'),
+      longDescription: messages.getMessage('flags.allLong'),
+    }),
```
shetzel, line 34: I didn't even know this flag existed on the legacy command.  What's the point?  Don't we default to all if no flags are passed?

src/commands/force/source/beta/pull.ts
```diff
@@ -0,0 +1,185 @@
...
+        files: successes.map((fileResponse) => fileResponse.filePath).filter(Boolean),
+      }),
+      this.tracking.updateRemoteTracking(
+        successes.map(({ state, fullName, type, filePath }) => ({ state, fullName, type, filePath })),
+        true // skip polling because it's a pull
+      ),
+    ]);
+  }
+
+  protected async retrieve(): Promise<void> {
```
WillieRuemmele, line 114: would it make sense to change this to `pull` instead? It's doing a little more than just a retrieve and might confuse the next person coming along. Same for push/deploy

src/commands/force/source/beta/status.ts
```diff
@@ -0,0 +1,169 @@
...
+      // keys like ApexClass__MyClass.cls
+      const conflictKeys = (await tracking.getConflicts()).map((conflict) => getKeyFromObject(conflict));
+      if (conflictKeys.length > 0) {
+        this.results = this.results.map((row) =>
+          conflictKeys.includes(getKeyFromStrings(row.type, row.fullName))
+            ? { ...row, state: `${row.state} (Conflict)` }
+            : row
+        );
+      }
+    }
```
shetzel, line 124: it seems like lines 76-124 could be wrapped into a STL function.  Is that possible?

src/commands/force/source/beta/status.ts
mshanemc: Yes.  I think it's a stupid flag and I'd love to get rid of it.

src/commands/force/source/beta/push.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    await Promise.all([
+      // Only fire the postdeploy event when we have results. I.e., not async.
+      this.deployResult ? this.lifecycle.emit('postdeploy', this.deployResult) : Promise.resolve(),
+      tracking.updateLocalTracking({
+        files: successNonDeletes.map((fileResponse) => fileResponse.filePath),
+        deletedFiles: successDeletes.map((fileResponse) => fileResponse.filePath),
+      }),
+      tracking.updateRemoteTracking(successes),
+    ]);
```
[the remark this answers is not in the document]
mshanemc: hmm.  I guess that could be a thing.  That's going to make a mess of source tracking, no matter what (conflicts!)

I'm changing to put it outside of the promises.

src/commands/force/source/beta/push.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  protected resolveSuccess(): void {
+    // there might not be a deployResult if we exited early with an empty componentSet
+    if (this.deployResult && this.deployResult.response.status !== RequestStatus.Succeeded) {
```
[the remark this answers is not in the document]
mshanemc: I have another one in progress for handling all the exit codes (partials, timeouts) but I need a replicable PartialSuccess repo.  Can we leave this for now?

src/commands/force/source/beta/pull.ts
mshanemc: translation: [I was too lazy to try to dig through toolbelt to find out and was hoping someone knew]

mshanemc: My reading is that it doesn't emit hooks...the postRetrieveHookInfo is only using the fileProperties from the mdapi response.  I think.

src/commands/force/source/beta/status.ts
mshanemc: done in https://github.com/salesforcecli/plugin-source/pull/258

src/commands/force/source/beta/push.ts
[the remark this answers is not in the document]
mshanemc: nope, I moved it since partials won't apply to push

PR #211, opened 2021-09-21
Warn users that CustomField requires CustomObject for source:retrieve

maggiben:
### What does this PR do?
Users are warned that they need to also retrieve the CustomObject when only CustomField is specified in the -m parameter
Users are warned that they need to also retrieve the CustomObject when only CustomField is specified in the package.xml supplied in the -x parameter

@W-7274674@


src/commands/force/source/retrieve.ts
```diff
@@ -131,4 +136,10 @@ export class Retrieve extends SourceCommand {
 
     return formatter.getJson();
   }
+
+  private wantsToRetrieveCustomFields(): boolean {
+    return this.componentSet.toArray().some((sourceComponent: SourceComponent) => {
```
mshanemc, line 141: error condition is

some  (type.name === Customfield && fullname === ComponentSet.WILDCARD)
! some ((type.name === CustomObject && fullname === ComponentSet.WILDCARD))

test/commands/source/retrieve.test.ts
```diff
@@ -266,4 +271,34 @@ describe('force:source:retrieve', () => {
     expect(displayStub.calledOnce).to.equal(false);
     expect(getJsonStub.calledOnce).to.equal(true);
   });
+
+  it('should warn users when retrieving CustomField with --metadata', async () => {
+    const metadata = 'CustomField:CustomObject__c.CustomField__c';
```
mshanemc, line 276: this -m should not produce the warning

`-m CustomField` should produce the warning

`-m CustomField,CustomObject` should not produce the warning

test/commands/source/retrieve.test.ts
```diff
@@ -266,4 +271,34 @@ describe('force:source:retrieve', () => {
...
+        return [exampleCustomFieldSourceComponent];
+      },
+    });
+    await runRetrieveCmd(['--metadata', metadata]);
+    expect(warnStub.calledOnce);
+    expect(warnStub.firstCall.firstArg).to.equal(messages.getMessage('wantsToRetrieveCustomFields'));
+  });
+
+  it('should warn users when retrieving CustomField with --manifest', async () => {
+    const manifest = 'package.xml';
```
mshanemc, line 291: package.xml should look something like a real package.xml.  Or a NUT might be easier (--json will have a warning property)

<types>
  <members>*
  <CustomField>
</types>

maggiben: Since we're interested in the `componentSet` array I tought It would be a good idea to stub `ComponentSetBuilder.build` with different reponses (given real examples) for both wildcard options as well as `CustomField,CustomObject` I think that should cover the cases where the warn should and should not be displayed.

src/commands/force/source/retrieve.ts
```diff
@@ -88,6 +88,11 @@ export class Retrieve extends SourceCommand {
       },
     });
 
+    const wantsToRetrieveCustomFields = this.wantsToRetrieveCustomFields();
+    if (wantsToRetrieveCustomFields) {
```
WillieRuemmele, line 92: could we simplify this to `if(this.wantsToRetrieveCustomFields())`?

src/commands/force/source/retrieve.ts
```diff
@@ -131,4 +136,14 @@ export class Retrieve extends SourceCommand {
 
     return formatter.getJson();
   }
+
+  private wantsToRetrieveCustomFields(): boolean {
+    const hasCustomField = this.componentSet.toArray().some((sourceComponent: SourceComponent) => {
```
WillieRuemmele, line 141: could you use `this.componentSet.has({fullName: ComponentSet.WILDCARD, type: 'CustomField'})` which might be more performant than iterating over all of them twice.

we only need to check this method if we're using `--metadata` or `--manifest` right? Can we only call this method when those params are present?

maggiben: @WillieRuemmele I agree we should only check on `-m` and `-x` but the `has` method on `componentSet` is `undefined`, so not sure that solution will work: 
image

maggiben: condition changed to `this.componentSet.has`

test/commands/source/retrieve.test.ts
```diff
@@ -266,4 +274,106 @@ describe('force:source:retrieve', () => {
...
+    const metadata = 'CustomField';
+    buildComponentSetStub.restore();
+    buildComponentSetStub = stubMethod(sandbox, ComponentSetBuilder, 'build').resolves({
+      retrieve: retrieveStub,
+      getPackageXml: () => packageXml,
+      toArray: () => {
+        return [exampleSourceComponent];
+      },
+      has: (component: ComponentLike) => {
+        expect(component).to.be.a.a('object').and.to.have.property('type');
```
WillieRuemmele, line 288: NIT: are there supposed to be two `a`'s? `to.be.a.a(`?

it's in the other tests too

WillieRuemmele (approved):

test/commands/source/retrieve.test.ts
maggiben: no there are not, yet the test passed o_0

PR #215, opened 2021-09-24
test: territory nut (weird folders use case)

mshanemc:
### What does this PR do?
Adds a nut for territory deploy/retrieve so we never regress those again.

original fixes were done on 
@W-9789567@
https://github.com/forcedotcom/cli/issues/1147


WillieRuemmele (approved):
:nuts: only took 2 mins to run all of these locally

PR #220, opened 2021-09-28
fix: remote-only deletes now supported

WillieRuemmele:
### What does this PR do?
`source:delete` would fail when trying to delete metadata that was only present in the org, and not locally

now it will delete remote-only metadata

@W-9892375@


src/commands/force/source/delete.ts
```diff
@@ -174,7 +176,7 @@ export class Delete extends DeployCommand {
 
   private deleteFilesLocally(): void {
     if (!this.getFlag('checkonly') && getString(this.deployResult, 'response.status') === 'Succeeded') {
-      this.sourceComponents.map((component) => {
+      this.componentSet.toArray().map((component: SourceComponent) => {
```
mshanemc, line 179: componentSet.toArray is going to have to iterate the whole thing (and it's a generator so it has to call next() on all of them) https://github.com/forcedotcom/source-deploy-retrieve/blob/aa79e2642a7aa35b05cb71c9f4cc0b0c21c21123/src/collections/lazyCollection.ts#L59

Perf idea: do that once and store it somewhere (ex: `this.componentSetAsArray`) to use in the various places.

mshanemc (approved):
approved.  I'm not particularly worried about perf on `delete` as much as the commands where I expected a lot more sourceMembers.

src/commands/force/source/delete.ts
```diff
@@ -174,7 +176,7 @@ export class Delete extends DeployCommand {
...
         // delete the content and/or the xml of the components
         if (component.content) {
           const stats = fs.lstatSync(component.content);
```
mshanemc, line 182: I'd also suggest thinking through if it's possible to parallelize these via Promise.all (ex: read all the files and build the file list of what needs to be deleted, then await all those deletes in a Promise.all).

I'm thinking of our windows fs pains.  Again, perhaps not a big deal for deletes.

shetzel (approved):

iowillhoit:
This is looking good @WillieRuemmele. QA'd this with the old plugin-source and then with your changes. Tried the following and all looks good: 
Remote only w/ prompts
Remote only, multiple deletes w/ prompts
Remote only no prompts
Remote only json w/ prompts
Remote only json no prompts

Retrieved w/ prompts
Retrieved, multiple deletes w/ prompts
Retrieved, no prompts
Retrieved, json w/ prompts
Retrieved, json no prompts

PR #231, opened 2021-09-29
feat: warning label for source:open

mshanemc:
What does this PR do?
adds security warning for the output of source:open (AT in URL)

What issues does this PR fix or reference?
@W-9953835@


shetzel (approved):
Approved, but NUTs are not passing reliably anymore and we need to address that.

cristiand391 (approved):

PR #265, opened 2021-10-26
Sm/status-codes

mshanemc:
### What does this PR do?
exit codes for `SucceededPartial` and the various incomplete statuses 
spinners with messages for retrieve

@W-9805781@


src/commands/force/source/beta/push.ts
```diff
@@ -89,13 +91,19 @@ export default class Push extends DeployCommand {
...
     // fire predeploy event for sync and async deploys
     await this.lifecycle.emit('predeploy', componentSet.toArray());
-    this.ux.log(`*** Pushing with ${this.isRest ? 'REST' : 'SOAP'} API v${componentSet.sourceApiVersion} ***`);
+
+    const isRest = await this.isRestDeploy();
+    this.ux.log(
+      `*** Pushing with ${isRest ? 'REST' : 'SOAP'}${
+        componentSet.sourceApiVersion ? ` API v${componentSet.sourceApiVersion}` : ''
+      } ***`
+    );
```
shetzel, line 100: I thought you were going to report both sourceApiVersion (when different from apiVersion) and apiVersion?  If you're going to report 1, then I would choose apiVersion.  Imagine if someone is overriding the apiVersion on the command line or by env var and the command reports the sourceApiVersion.  That would be confusing.

shetzel (changes requested):

src/commands/force/source/beta/push.ts
mshanemc: OK, here's that for push and deploy.

shetzel (approved):

shetzel (approved):

PR #291, opened 2021-11-08
fix: handle a corrupt stash.json file

shetzel:
### What does this PR do?
When a JsonParseError is thrown while reading `stash.json` for the deploy ID, the error is now caught, the file is renamed to stash.json_corrupted_<timestamp> and an error message is displayed that shows the parse failure and the location of the renamed stash.json so users can inspect the file, or possibly rebuild a valid file from the corrupt version.

@W-7837761@

Testing:
1. change stash.json to be invalid json
2. delete stash.json
3. change stash.json to be empty

In all cases an error should be thrown.  In 1 & 3 the stash.json file will be renamed so running the command again will not result in the same behaviour.


mshanemc (approved):
it's a fine solution for a rare problem

PR #294, opened 2021-11-11
Sm/ts updates

mshanemc:
### What does this PR do?
fix ts-update problems

testing notes:
I regularly see the `sfdx typescript:update` command throw errors about `??=`.  You can replicate that by updating tsnode to latest (10.4.0 is the one I see the problem on).

It compiles (tsc) just fine, and the `test-compile` does fine, too.  So this is some bug with the ts-node test running stuff.

@W-10054614@

### DIFF
--- a/test/formatters/deployResultFormatter.test.ts
+++ b/test/formatters/deployResultFormatter.test.ts
@@ -27,6 +27,7 @@ describe('DeployResultFormatter', () => {
 
   const logger = Logger.childFromRoot('deployTestLogger').useMemoryLogging();
   let ux;
+
   let logStub: sinon.SinonStub;
   let styledHeaderStub: sinon.SinonStub;
   let tableStub: sinon.SinonStub;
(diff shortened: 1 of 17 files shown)


test/formatters/deployResultFormatter.test.ts
```diff
@@ -58,7 +59,7 @@ describe('DeployResultFormatter', () => {
     it('should return expected json for a success', async () => {
       const deployResponse = JSON.parse(JSON.stringify(deployResultSuccess.response)) as DeployCommandResult;
       const expectedSuccessResults = deployResultSuccess.response as DeployCommandResult;
-      const formatter = new DeployResultFormatter(logger, ux, {}, deployResultSuccess);
+      const formatter = new DeployResultFormatter(logger, ux as UX, {}, deployResultSuccess);
```
WillieRuemmele, line 62: we can't DRY this up because ux is type `stubInterface` and not a real `UX` correct?

WillieRuemmele (approved):

shetzel (approved):

PR #301, opened 2021-11-15
fix: add mdapi:deploy:cancel command, refactor base classes to support MDAPI and SOURCE stash keys

WillieRuemmele:
### What does this PR do?
moves `mdapi:deploy:cancel` command and messages
extends `plugin-source` base classes correctly
updates base classes to support `MDAPI_DEPLOY` stash key
adds UTs
adds NUTs

`source:deploy:cancel`,`source:deploy:report`, `mdapi:deploy:cancel` all validate the passed in deploy ID parameter
fixes issue when stash is missing a requested key/value

@W-9603715@

### DIFF
--- a/src/commands/force/mdapi/deploy/cancel.ts
+++ b/src/commands/force/mdapi/deploy/cancel.ts
@@ -9,7 +9,6 @@ import * as os from 'os';
 import { flags, FlagsConfig } from '@salesforce/command';
 import { Messages } from '@salesforce/core';
 import { Duration } from '@salesforce/kit';
-import { getString } from '@salesforce/ts-types';
 import { RequestStatus } from '@salesforce/source-deploy-retrieve';
 import { DeployCommand } from '../../../../deployCommand';
 import {
@@ -56,7 +55,7 @@ export class Cancel extends DeployCommand {
   }
 
   protected resolveSuccess(): void {
-    const status = getString(this.deployResult, 'response.status');
+    const status = this.deployResult.response.status;
     if (status !== RequestStatus.Canceled) {
       this.setExitCode(1);
     }
--- a/src/commands/force/mdapi/deploy/cancel.ts
+++ b/src/commands/force/mdapi/deploy/cancel.ts
@@ -18,7 +18,7 @@ import {
 } from '../../../../formatters/deployCancelResultFormatter';
 
 Messages.importMessagesDirectory(__dirname);
-const messages = Messages.loadMessages('@salesforce/plugin-source', 'cancel');
+const messages = Messages.loadMessages('@salesforce/plugin-source', 'md.cancel');
 
 export class Cancel extends DeployCommand {
   public static readonly description = messages.getMessage('description');
@@ -30,13 +30,15 @@ export class Cancel extends DeployCommand {
       default: Duration.minutes(DeployCommand.DEFAULT_WAIT_MINUTES),
       min: Duration.minutes(1),
       description: messages.getMessage('flags.wait'),
-      longDescription: messages.getMessage('flagsLong.wait'),
+      longDescription: messages.getMessage('flags.waitLong'),
     }),
     jobid: flags.id({
       char: 'i',
       description: messages.getMessage('flags.jobid'),
     }),
   };
+  // The most important difference between this and source:deploy:cancel
+  public isSourceStash = false;
 
   public async run(): Promise<DeployCancelCommandResult> {
     await this.cancel();
--- a/src/commands/force/mdapi/deploy/cancel.ts
+++ b/src/commands/force/mdapi/deploy/cancel.ts
@@ -7,7 +7,7 @@
 
 import * as os from 'os';
 import { flags, FlagsConfig } from '@salesforce/command';
-import { Messages } from '@salesforce/core';
+import { Messages, SfdxError } from '@salesforce/core';
 import { Duration } from '@salesforce/kit';
 import { RequestStatus } from '@salesforce/source-deploy-retrieve';
 import { DeployCommand } from '../../../../deployCommand';
@@ -34,6 +34,13 @@ export class Cancel extends DeployCommand {
     jobid: flags.id({
       char: 'i',
       description: messages.getMessage('flags.jobid'),
+      validate: (val) => {
+        if (val.startsWith('0Af')) {
+          return true;
+        } else {
+          throw SfdxError.create('@salesforce/plugin-source', 'deploy', 'invalidDeployId');
+        }
+      },
     }),
   };
   // The most important difference between this and source:deploy:cancel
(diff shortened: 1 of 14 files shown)


src/commands/force/mdapi/deploy/cancel.ts
```diff
@@ -0,0 +1,72 @@
...
+    const deployId = this.resolveDeployId(this.getFlag<string>('jobid'));
+
+    const deploy = this.createDeploy(deployId);
+    await deploy.cancel();
+
+    this.deployResult = await this.poll(deployId);
+  }
+
+  protected resolveSuccess(): void {
+    const status = getString(this.deployResult, 'response.status');
```
mshanemc, line 59: let's avoid `getString` where we have good types.
Where we don't, let's make good types.

src/deployCommand.ts
```diff
@@ -18,17 +23,19 @@ interface StashFile {
...
       this.ux.log(`Deploy ID: ${id}`);
     }
   });
-
+  // used to determine the correct stash.json key
+  protected isSourceStash = true;
   protected deployResult: DeployResult;
   /**
-   * Request a report of an in-progess or completed deployment.
+   * Request a report of an in-progress or completed deployment.
```
mshanemc, line 38: +1

src/deployCommand.ts
```diff
@@ -47,18 +54,39 @@ export abstract class DeployCommand extends SourceCommand {
...
+  protected resolveDeployId(id?: string): string {
     let stash: ConfigFile<StashFile>;
     if (id) {
       return id;
     } else {
       try {
         stash = this.getStash();
         stash.readSync(true);
-        const deployId = asString((stash.get(DeployCommand.STASH_KEY) as { jobid: string }).jobid);
+        const deployId = asString(
```
mshanemc, line 83: why does this need `asString`?  there's already a type assertion inside it.

[I'd prefer to avoid these non-standard TS "helpers" wherever we can to keep things more approachable for externals]

src/sourceCommand.ts
```diff
@@ -20,7 +20,7 @@ export type ProgressBar = {
 };
 
 export abstract class SourceCommand extends SfdxCommand {
-  public static readonly DEFAULT_SRC_WAIT_MINUTES = 33;
+  public static readonly DEFAULT_WAIT_MINUTES = 33;
```
mshanemc, line 23: is the idea that the `mdapi:deploy` command will just have it's own flag that's set to zero (that is, just not using the default)?

test/nuts/mdapi.nut.ts
```diff
@@ -58,6 +60,38 @@ describe('mdapi NUTs', () => {
     });
   });
 
+  describe('mdapi:deploy:cancel', () => {
+    it('will cancel an mdapi deploy via the stash.json', () => {
+      execCmd('force:source:convert --outputdir mdapi');
+      // TODO: once mdapi:deploy is migrated switch to execCmd
```
mshanemc, line 66: +1 for the TODO

test/nuts/mdapi.nut.ts
```diff
@@ -58,6 +60,38 @@ describe('mdapi NUTs', () => {
...
+      const deploy = JSON.parse(exec('sfdx force:mdapi:deploy -d mdapi -w 0 --json', { silent: true })) as {
+        result: { id: string };
+      };
+      const result = execCmd(`force:mdapi:deploy:cancel --json --jobid ${deploy.result.id}`);
+      expect(result.jsonOutput.status).to.equal(0);
+      const json = result.jsonOutput.result;
+      expect(json).to.have.property('canceledBy');
+      expect(json).to.have.property('status');
+      expect(getString(json, 'status')).to.equal('Canceled');
```
mshanemc, line 90: this would be better if the `cancel` command's output type were imported and used `execCmd<x>`

src/commands/force/mdapi/deploy/cancel.ts
```diff
@@ -0,0 +1,72 @@
...
+    }),
+    jobid: flags.id({
+      char: 'i',
+      description: messages.getMessage('flags.jobid'),
+    }),
+  };
+  // The most important difference between this and source:deploy:cancel
+  public isSourceStash = false;
+
+  public async run(): Promise<DeployCancelCommandResult> {
```
mshanemc, line 43: oh, so both cancel commands have the same output shape?  That's great!

test/nuts/mdapi.nut.ts
```diff
@@ -58,6 +60,38 @@ describe('mdapi NUTs', () => {
...
   });
 
+  describe('mdapi:deploy:cancel', () => {
+    it('will cancel an mdapi deploy via the stash.json', () => {
+      execCmd('force:source:convert --outputdir mdapi');
+      // TODO: once mdapi:deploy is migrated switch to execCmd
+      const deploy = JSON.parse(exec('sfdx force:mdapi:deploy -d mdapi -w 0 --json', { silent: true })) as {
+        result: { id: string };
+      };
+      const result = execCmd('force:mdapi:deploy:cancel --json');
```
mshanemc, line 70: Is there any chance this won't cancel in time (I think I remember that with other NUTs where sometimes it would succeed before the cancel request arrived, depending on how busy the mdapi was).

mshanemc (approved):
approved with non-blocking suggestions/questions

src/sourceCommand.ts
WillieRuemmele: hmmm good point. for `mdapi:deploy:cancel` in toolbelt it was using the `source:deploy` default, it seemed weird to have this variable mention `SRC` when it's now used in both `source` and `mdapi`

very open to reverting this back

src/commands/force/mdapi/deploy/cancel.ts
WillieRuemmele: yeah - they're identical except for messages and the `stash.json` key it uses

src/sourceCommand.ts
mshanemc: I think it's good since 33 seems to be most everywhere.  Just making sure we weren't gonna come back and put a `DEFAULT_MDAPI_WAIT_MINUTES` or whatever.

mshanemc:
QA notes:

👎🏻  run --help
There's some odd indentation stuff happening.  Compare `./bin/run force:mdapi:deploy:cancel --help` to `./bin/run force:source:deploy:cancel --help` to see the difference

✅  run without a username (good error)
❌ run without a jobid on an org with no stash yields 
```
ERROR running force:mdapi:deploy:cancel:  Cannot read property 'jobid' of undefined
```
vs toolbelt
```
ERROR running force:mdapi:deploy:cancel:  A required parameter [jobid] is missing.
```

👎🏻  ask to cancel a jobId that doesn't exist.  Toolbelt is similarly bad.
```
ERROR running force:mdapi:deploy:cancel:  INVALID_CROSS_REFERENCE_KEY: invalid cross reference id
```

✅  cancel a job that already succeeded fails with `ERROR running force:mdapi:deploy:cancel:  The cancel command failed due to: INVALID_ID_FIELD: Deployment already completed`

👎🏻  cancel a jobid that is a CustomObject's id fails with `ERROR running force:mdapi:deploy:cancel:  The cancel command failed due to: UNKNOWN_EXCEPTION: unexpected metadata` [yet we know what the prefix has to be `0Af` and could validate the ID is of the correct type before even making an API call)

👎🏻  cancel a job, then cancel it again fails with `ERROR running force:mdapi:deploy:cancel:  INVALID_ID_FIELD: Deployment already completed` (kinda true, I guess?)
 - 👎🏻   json output of toolbelt had a nicer named error
  ```
       "name": "CancelFailed",
       "message": "The cancel command failed due to: INVALID_ID_FIELD: Deployment already completed",
  ``` 
  that's changed to
  ```
        "name": "sf:INVALID_ID_FIELD",
        "message": "INVALID_ID_FIELD: Deployment already completed",     
  ```

✅  delete `stash.json` and then run without an id fails with `ERROR running force:mdapi:deploy:cancel:  No deploy ID was provided or found in deploy history`
✅  json output of successful cancel matches between toolbelt and plugin

PR #302, opened 2021-11-16
Sh/mdapi convert cmd

shetzel:
### What does this PR do?
Adds the `force:mdapi:beta:convert` command, which has parity with `force:mdapi:convert` from the salesforce-alm plugin.

@W-9603700@

Testing Notes:
1. use the command to convert all metadata of several repos
2. use all command flags
3. point to non-existant files to test error output
4. test converting metadata that doesn't match flag filters
5. test with forceignore
6. test json and non-json output
7. compare all output with `force:mdapi:convert`


messages/md.convert.json
```diff
@@ -0,0 +1,28 @@
...
+  "flagsLong": {
+    "rootdir": "The root directory that contains the metadata you retrieved using Metadata API.",
+    "outputdir": "The directory to store your files in after they’re converted to the source format. Can be an absolute or relative path.",
+    "manifest": [
+      "The complete path to the manifest (package.xml) file that specifies the metadata types to convert.",
+      "If you specify this parameter, don’t specify --metadata or --sourcepath."
+    ],
+    "metadatapath": [
+      "A comma-separated list of paths to the local metadata files to convert. The supplied paths can be to a single file (in which case the operation is applied to only one file) or to a folder (in which case the operation is applied to all metadata types in the directory and its sub-directories).",
+      "If you specify this parameter, don’t specify --manifest or --metadata."
```
jshackell-sfdc, line 23:
```suggestion
      "If you specify this parameter, don’t specify --manifest or --metadata. If the comma-separated list you’re supplying contains spaces, enclose the entire comma-separated list in one set of double quotes."
```

jshackell-sfdc: Added back one sentence that was in legacy command help.  I assume it's still correct?

messages/md.convert.json
```diff
@@ -0,0 +1,28 @@
+{
+  "description": "convert metadata from the Metadata API format into the source format\nConverts metadata retrieved via Metadata API into the source format used in Salesforce DX projects.\nTo use Salesforce CLI to work with components that you retrieved via Metadata API, first convert your files from the metadata format to the source format using \"sfdx force:mdapi:convert\".\n\nTo convert files from the source format back to the metadata format, so that you can deploy them using \"sfdx force:mdapi:deploy\", run \"sfdx force:source:convert\".",
```
jshackell-sfdc, line 2:
```suggestion
  "description": "convert metadata from the Metadata API format into the source format\nConverts metadata retrieved via Metadata API into the source format used in Salesforce DX projects.\n\nTo use Salesforce CLI to work with components that you retrieved via Metadata API, first convert your files from the metadata format to the source format using \"sfdx force:mdapi:convert\".\n\nTo convert files from the source format back to the metadata format, so that you can deploy them using \"sfdx force:mdapi:deploy\", run \"sfdx force:source:convert\".",
```

src/commands/force/mdapi/beta/convert.ts
```diff
@@ -0,0 +1,215 @@
...
+        if (throwOnENOENT) {
+          throw SfdxError.create('@salesforce/plugin-source', 'md.convert', 'InvalidFlagPath', [flagName, path]);
+        }
+        fs.mkdirSync(path, { recursive: true });
+      }
+    }
+  }
+
+  private resolveRootDir(rootDir: string): string {
+    const resolvedRootDir = resolve(rootDir.trim());
```
mshanemc, line 153: why not do the trimming either in `ensureFlagPath` or in the flags themselves via `flag.parse` instead of in each of these?

src/commands/force/mdapi/beta/convert.ts
```diff
@@ -0,0 +1,215 @@
...
+    let paths: string[];
+    if (metadatapath) {
+      paths = metadatapath;
+    } else if (!manifest && !metadata) {
+      paths = [this.rootDir];
+    }
+
+    this.componentSet = await ComponentSetBuilder.build({
+      sourcepath: paths,
+      manifest: manifest && {
```
mshanemc, line 91: I didn't know that did that.

src/commands/force/mdapi/beta/convert.ts
```diff
@@ -0,0 +1,215 @@
...
+        manifestPath: manifest,
+        directoryPaths: [this.rootDir],
+      },
+      metadata: metadata && {
+        metadataEntries: metadata,
+        directoryPaths: [this.rootDir],
+      },
+    });
+
+    const numOfComponents = this.componentSet.getSourceComponents().toArray().length;
```
mshanemc, line 101: why not this.componentSet.size?

test/nuts/mdapi.nut.ts
```diff
@@ -58,6 +74,58 @@ describe('mdapi NUTs', () => {
     });
   });
 
+  describe('mdapi:beta:convert', () => {
```
mshanemc, line 77: Now that we have both convert commands in here, I'd be more interested in a round-trip conversion NUT on some decently-complex repo, with an assertion that everything matched.

Since there's no org involved, it'd run pretty fast.

mshanemc (approved):

src/commands/force/mdapi/beta/convert.ts
shetzel: Because it would report a size of 1 with certain kinds of ComponentSets (built from metadata or manifest) while there were no source components.  This way is more reliable when you care about the number of source components.

test/nuts/mdapi.nut.ts
shetzel: Yeah, we could do that.  I know the ebikes-lwc source is not in the expected format in the repo (due to 1 space in the xml header) but maybe dreamhouse is.  Regardless, we could source:convert, mdapi:convert, then source:convert again and compare.

src/commands/force/mdapi/beta/convert.ts
shetzel: Yeah that's a good point.  I think either place could work.  I'll work on that.

src/commands/force/mdapi/beta/convert.ts
mshanemc: Your PRs are so educational.  Let me check my understanding because that is 100% not intuitive to me....

`.size` only gives the number of _source-backed-components_ 
but `.getSourceComponents` will give you an array that includes _non-source-backed-components_, too?

mshanemc (approved):
even more approve!

mshanemc:
QA notes
test-repo https://github.com/SalesforceFoundation/EDA (source:convert `force-app` to `/src`)
✅  `mdapi:convert` it back ( -d -r )
✅ `sfdx force:mdapi:beta:convert -d source2 -r unpackaged -p unpackaged/pre` (was already in mdapi format...contains field/object translations)
✅ multiple -p with unpackaged/[pre,post,config].  Allows illegal object/fieldnames where cumulusCI has placeholders, like `source2/main/default/layouts/___NAMESPACE___Course_Offering__c-___NAMESPACE___HEDA Course Offering Layout.layout-meta.xml`.  Probably a good thing

    ❌  if some mdapi-format folder [includes a file with a `-meta.xml` suffix](https://github.com/SalesforceFoundation/EDA/blob/main/unpackaged/config/settings_product/customMetadata/___NAMESPACE___Product_Registry.TALDemoProductInformation.md-meta.xml), then convert 
    1. duplicates the suffix `md-meta.xml-meta.xml`
    2. misclassifies it as an `EmailServicesFunction` because of the `.xml` ending.
    
```text
    Add    ___NAMESPACE___Product_Registry.TALDemoProductInformation.md-meta                   
  EmailServicesFunction    source2/main/default/customMetadata/___NAMESPACE___Product_Registry.TALDemoProductInformation.md-meta.xml-meta.xml
```

Toolbelt just ignored that file.  

[digging deeper, took out the `___NAMESPACE___` stuff].  Toolbelt still ignores, and plugin-source still creates the duplicate file.  

> Suggestion: either ignore files ending in `-meta.xml` (consistency) OR "fix" them by removing the suffix when determining metadata type and new filename from SDR.
  
✅ deep folder "hoisted" to top of source outputDir using -x `sfdx force:mdapi:beta:convert -d source3 -x unpackaged/test_packaged/primary_affiliations/package.xml -r unpackaged/test_packaged/primary_affiliations`

👎🏻 --rootdir with a filepath results in `ERROR running force:mdapi:beta:convert:  The rootdir command parameter specifies an invalid path: unpackaged/config/settings_product/package.xml` which is harder to grok than, "your `--rootdir` is supposed to be a directory"

❌ I renamed an apex class to `/Users/shane.mclaughlin/eng/repros/EDA/unpackaged/config/settings_product/classes/TALDemoSettingsProductInfoAPIService.clsx-meta.xml` (note `clsx`) which yields
```
Add    TALDemoSettingsProductInfoAPIService.clsx-meta      EmailServicesFunction  source4/main/default/emailservices/TALDemoSettingsProductInfoAPIService.clsx-meta.xml-meta.xml
``` 
I expected that to error about the file.  Toolbelt errors with `ERROR running force:mdapi:convert:  Expected metadata file at path: /Users/shane.mclaughlin/eng/repros/EDA/unpackaged/config/settings_product/classes/TALDemoSettingsProductInfoAPIService.cls-meta.xml
` (it saw the .cls file and couldn't find the corresponding file)

[I know this isn't ideal use case--the code is already in source format.  There's just a few scenarios where that's not being handled correctly when the types aren't easily identifiable.]

PR #305, opened 2021-11-18
chore: bump deps

WillieRuemmele:
### What does this PR do?
bumps core,SDR
@W-0@

### DIFF
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -2,6 +2,12 @@
   "extends": "@salesforce/dev-config/tsconfig-test",
   "include": ["./**/*.ts"],
   "compilerOptions": {
-    "skipLibCheck": true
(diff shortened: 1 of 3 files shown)


test/tsconfig.json
```diff
@@ -2,6 +2,12 @@
   "extends": "@salesforce/dev-config/tsconfig-test",
   "include": ["./**/*.ts"],
   "compilerOptions": {
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "baseUrl": "..",
```
mshanemc, line 6: will compile using the top-level imports of this repo (when they satisfy dependencies) instead of transient dependencies' versions.

This matches the top-level `tsconfig` so we don't have "project builds but tests don't compile"

PR #309, opened 2021-11-20
fix: conditionally include folder

shetzel:
### What does this PR do?
only include the folder ListMetadataQuery for jsForce's connection.metadata.list when it's a folder type.  Even an undefined folder value causes problems for the ListView type.

@W-10203090@

### DIFF
--- a/src/commands/force/mdapi/listmetadata.ts
+++ b/src/commands/force/mdapi/listmetadata.ts
@@ -68,7 +68,10 @@ export class ListMetadata extends SourceCommand {
 
     this.validateResultFile();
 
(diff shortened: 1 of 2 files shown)


WillieRuemmele (approved):

mshanemc (approved):

iowillhoit:
QA: Confirmed that ListView meta data was not displayed before this fix and is displaying after the fix. Also listed a few other metadata types post-update and everything appears to be fine 👍 

Screen Shot 2021-11-23 at 10 41 59 AM

Screen Shot 2021-11-23 at 10 42 17 AM

Screen Shot 2021-11-23 at 10 45 47 AM

PR #313, opened 2021-11-24
Wr/bundle helper delete

WillieRuemmele:
### What does this PR do?
allows users to `source:delete` an individual bundle file by deleting the file, and then performing a deploy on the "new" bundle

@W-10184606@

### DIFF
--- a/src/commands/force/source/delete.ts
+++ b/src/commands/force/source/delete.ts
@@ -277,6 +277,7 @@ export class Delete extends DeployCommand {
       type: bundle.type.name,
       filePath: sourcepath,
     });
+    // stash the file in case we need to restore it due to failed deploy/aborted command
     this.stashPath = path.join(os.tmpdir(), 'source_delete', fileName);
     fs.mkdirSync(path.dirname(this.stashPath), { recursive: true });
     fs.copyFileSync(sourcepath, this.stashPath);
--- a/src/commands/force/source/delete.ts
+++ b/src/commands/force/source/delete.ts
@@ -6,12 +6,15 @@
  */
 import * as os from 'os';
 import * as fs from 'fs';
+import * as path from 'path';
 import { confirm } from 'cli-ux/lib/prompt';
 import { flags, FlagsConfig } from '@salesforce/command';
 import { Messages } from '@salesforce/core';
 import {
   ComponentSet,
+  ComponentStatus,
   DestructiveChangesType,
+  FileResponse,
   MetadataComponent,
   RequestStatus,
   SourceComponent,
@@ -20,7 +23,7 @@ import { Duration, env, once } from '@salesforce/kit';
 import { getString } from '@salesforce/ts-types';
 import { DeployCommand } from '../../../deployCommand';
 import { ComponentSetBuilder } from '../../../componentSetBuilder';
-import { DeployCommandResult } from '../../../formatters/deployResultFormatter';
+import { DeployCommandResult, DeployResultFormatter } from '../../../formatters/deployResultFormatter';
 import { DeleteResultFormatter } from '../../../formatters/deleteResultFormatter';
 import { ProgressFormatter } from '../../../formatters/progressFormatter';
 import { DeployProgressBarFormatter } from '../../../formatters/deployProgressBarFormatter';
@@ -79,9 +82,12 @@ export class Delete extends DeployCommand {
   protected xorFlags = ['metadata', 'sourcepath'];
   protected readonly lifecycleEventNames = ['predeploy', 'postdeploy'];
   private isRest = false;
-  private deleteResultFormatter: DeleteResultFormatter;
+  private deleteResultFormatter: DeleteResultFormatter | DeployResultFormatter;
   private aborted = false;
   private components: MetadataComponent[];
+  // create the delete FileResponse as we're parsing the comp. set to use in the output
+  private mixedDeployDelete: { deploy: string[]; delete: FileResponse[] } = { delete: [], deploy: [] };
+  private stashPath: string;
 
   private updateDeployId = once((id: string) => {
     this.displayDeployId(id);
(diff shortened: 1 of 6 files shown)


src/commands/force/source/delete.ts
```diff
@@ -133,6 +142,23 @@ export class Delete extends DeployCommand {
...
+    if (sourcepaths) {
+      // determine if user is trying to delete a single file from a bundle, which is actually just an fs delete operation
+      // and then a constructive deploy on the "new" bundle
+      this.components
+        .filter((comp) => {
+          if (comp.type.strategies?.adapter === 'bundle' && comp instanceof SourceComponent) return true;
+        })
+        .map((bundle: SourceComponent) => {
+          sourcepaths.map((sourcepath) => {
+            // walkContent returns absolute paths while sourcepath will usually be relative
```
mshanemc, line 154: test case for NUTs: delete an entire bundle (sourcepath points to `lwc/someCmp`) does a delete.

src/commands/force/source/delete.ts
```diff
@@ -170,46 +196,121 @@ export class Delete extends DeployCommand {
     if (status !== RequestStatus.Succeeded && !this.aborted) {
       this.setExitCode(1);
     }
+    // if deploy failed OR the operation was cancelled, restore the stashed files if they exist
+    else if (status !== RequestStatus.Succeeded || this.aborted) {
+      this.mixedDeployDelete.delete.map((file) => {
+        this.restoreFileFromStash(file.filePath);
```
mshanemc, line 202: this could also be Promise.all if the stash operations were async?

src/commands/force/source/delete.ts
```diff
@@ -170,46 +196,121 @@ export class Delete extends DeployCommand {
...
+        },
+      } as unknown as DeployCommandResult;
+    }
+
     return this.deleteResultFormatter.getJson();
   }
 
   private deleteFilesLocally(): void {
     if (!this.getFlag('checkonly') && getString(this.deployResult, 'response.status') === 'Succeeded') {
       this.components.map((component: SourceComponent) => {
```
mshanemc, line 251: also a good place for Promise.all

src/commands/force/source/delete.ts
```diff
@@ -221,6 +322,14 @@ export class Delete extends DeployCommand {
         }
       });
 
+      if (this.mixedDeployDelete.delete.length) {
```
mshanemc, line 325: if we're going go overwrite `local` with the `mixedDeployDelete` value, do we need to do line 318?

test/nuts/delete.nut.ts
```diff
@@ -12,6 +12,16 @@ import { expect } from 'chai';
 import { execCmd } from '@salesforce/cli-plugins-testkit';
 import { SourceTestkit } from '@salesforce/source-testkit';
 import { exec } from 'shelljs';
+import { FileResponse } from '@salesforce/source-deploy-retrieve';
+
+const queryOrgAndFS = (name: string, fsPath: string): void => {
+  const cmd = `force:data:soql:query -q "SELECT IsNameObsolete FROM SourceMember WHERE MemberType='LightningComponentBundle' AND MemberName='${name}'" -t --json`;
```
mshanemc, line 18: suggestion: get a connection from the setup org:create step, similar to what you did in the `deployDestructive.nut` 

```ts
conn = await Connection.create({
      authInfo: await AuthInfo.create({
        username: (session.setup[0] as { result: { username: string } }).result?.username,
      }),
    });
```

so it doesn't create another cross-plugin dependency

mshanemc (changes requested):
most comments are style/perf related and optional (I don't think delete will ever have as many SrcCmp as deploy/retrieve so I'm less worried about parallelizing fs ops).  

The only blocking request for change is the one about deleting an entire bundle in the NUT to make sure that works as expected.

src/commands/force/source/delete.ts
```diff
@@ -191,20 +190,22 @@ export class Delete extends DeployCommand {
   /**
    * Checks the response status to determine whether the delete was successful.
    */
-  protected resolveSuccess(): void {
-    const status = getString(this.deployResult, 'response.status');
+  protected async resolveSuccess(): Promise<void> {
+    const status = this.deployResult?.response?.status ?? undefined;
```
mshanemc, line 194: ?? means shortcut to undefined if anything in the chain is undefined/null
```suggestion
    const status = this.deployResult?.response?.status;
```

src/commands/force/source/delete.ts
```diff
@@ -191,20 +190,22 @@ export class Delete extends DeployCommand {
...
     }
     // if deploy failed OR the operation was cancelled, restore the stashed files if they exist
     else if (status !== RequestStatus.Succeeded || this.aborted) {
-      this.mixedDeployDelete.delete.map((file) => {
-        this.restoreFileFromStash(file.filePath);
-      });
+      await Promise.all(
+        this.mixedDeployDelete.delete.map((file) => {
+          return this.restoreFileFromStash(file.filePath);
+        })
```
mshanemc, line 203:
```suggestion
        this.mixedDeployDelete.delete.map((file) => this.restoreFileFromStash(file.filePath))
```

src/commands/force/source/delete.ts
```diff
@@ -146,14 +147,12 @@ export class Delete extends DeployCommand {
       // determine if user is trying to delete a single file from a bundle, which is actually just an fs delete operation
       // and then a constructive deploy on the "new" bundle
       this.components
-        .filter((comp) => {
-          if (comp.type.strategies?.adapter === 'bundle' && comp instanceof SourceComponent) return true;
-        })
+        .filter((comp) => comp.type.strategies?.adapter === 'bundle' && comp instanceof SourceComponent)
         .map((bundle: SourceComponent) => {
-          sourcepaths.map((sourcepath) => {
+          sourcepaths.map(async (sourcepath) => {
```
mshanemc, line 152: does that actually work?  Seems like there should be a Promise.all outside of it

WillieRuemmele: I haven't seen any issues from this... but I agree 🤔

mshanemc (approved):

mshanemc:
QA Notes:
- using dreamhouse

✅ provides warning before doing the delete (using -p).  exits with (n) and deletes with (y)
✅ file was deleted from the org
❌  output shows three deployed files but not a deleted file.  json has the same problems
✅ gives decent error when you point at a file that doesn't exist
✅ warns about all file when deleting the whole lwc bundle
✅ when delete fails, component is *not* deleted
✅ 😮 delete file from org, undelete locally (git) [so that it only exists locally] then source:delete it again

PR #347, opened 2021-12-03
feat: add deploy filter for sourcepaths

jayree:
### What does this PR do?

Adds a `filter` parameter for `source:deploy` which can be used together with `sourcepaths` 

Before `source-deploy-retrieve` the `source:deploy` command was able to deploy each `packageDir` one by one in the order specified in the `sfdx-project.json`, if you used `manifest` as a parameter.

This was a way to deploy only selected components from each `packageDir` with a manifest file, which contains all selected components from all folders.

As this behaviour is not possible with `source-deploy-retrieve` anymore, a parameter `filter` togehter with `sourcepaths` can bring back this "old" behaviour. Iteration of the folders needs to be done manually.

### DIFF
--- a/command-snapshot.json
+++ b/command-snapshot.json
@@ -72,6 +72,7 @@
     "flags": [
       "apiversion",
       "checkonly",
+      "filter",
       "ignoreerrors",
       "ignorewarnings",
       "json",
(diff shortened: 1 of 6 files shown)


shetzel:
After discussing this with Shane and Jochen, we've decided that it's preferable to use `sfdx force:source:deploy` to target "troublesome" metadata types that need to be deployed before the rest of a project.

PR #355, opened 2021-12-21
Sm/deploy-consolidation

mshanemc:
### What does this PR do?
refactoring redundant code, especially on deploy
- validating jobIds
- use oclif exactlyOne instad of xorFlags
-  hosting duplicate stuff from children of deployCommand to deployCommand
- gets stash key from command it (and has to manually set it in UT)

@W-9603711@


WillieRuemmele (approved):

src/functions.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+import { SfdxError } from '@salesforce/core';
+
+export const isValidDeployId = (id: string): boolean => {
+  if (id.startsWith('0Af')) {
+    return true;
+  } else {
+    throw SfdxError.create('@salesforce/plugin-source', 'deploy', 'invalidDeployId');
+  }
+};
```
shetzel, line 15: Rather than having this in its own file it could be a static function on DeployCommand.

shetzel (approved):
nothing blocking, but I'd definitely prefer a static function on DeployCommand to a new standalone functions file.

PR #360, opened 2021-12-29
fix: add variable polling times based on comp. set

WillieRuemmele:
### What does this PR do?
adds variable polling frequency to deploy/push/pull/retrieve

I ended up creating a piece-wise function to capture "small" vs "medium" vs "large" vs "larger" deploys - I'm very open to changing the boundaries and frequencies each should return. I did notice a difference between 250ms and 100ms

@W-9586672@

requires https://github.com/forcedotcom/source-deploy-retrieve/pull/520

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@oclif/config": "^1.18.1",
     "@salesforce/command": "^4.2.0",
     "@salesforce/core": "^2.31.0",
-    "@salesforce/source-deploy-retrieve": "^5.8.0",
+    "@salesforce/source-deploy-retrieve": "^5.8.4",
     "@salesforce/source-tracking": "^0.5.1",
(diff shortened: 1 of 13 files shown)


peternhale:
@WillieRuemmele not sure if this helps, but the underlying module [ts-retry-promise](https://www.npmjs.com/package/ts-retry-promise#configuration) has configuration options that allow one to specify a non-linear delays.

mshanemc (commented):
I'm not opposed to this being in plugin-source.  Thinking of `sf` and vscode and other potential users...
would it make more sense to put this in SDR since it would have a component set already?  Then if the user didn't specify some polling interval, your calculated intervals would be the default behavior?

mshanemc (approved):

PR #380, opened 2022-01-07
test: nuts for lightning email templates

mshanemc:
### What does this PR do?
brings folder-type nuts test repo into plugin-source
adds examples of EmailTemplateFolder

This requires no other PRs to merge in SDR or STL

it's somewhat related to @W-10385112@ but should give us a heads up if anything else weird starts to happen with EmailTemplateFolder type where it's not compatible with EmailFolder

### DIFF
--- a/test/nuts/folderTypes.nut.ts
+++ b/test/nuts/folderTypes.nut.ts
@@ -18,7 +18,7 @@ describe('metadata types that go in folders', () => {
   before(async () => {
     session = await TestSession.create({
       project: {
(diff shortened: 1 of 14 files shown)


shetzel (approved):
I have mixed emotions about creating projects like this in the source plugin repo.  This could easily explode to many, many projects.  Then again, having a repo for each seems like a bad idea too.  Maybe this is fine for now and if it starts to get out of hand we can think about alternatives.

mshanemc:
closing it while I make a proper solution for EF / ETF disambiguation

PR #383, opened 2022-01-10
Sh/mdapi retrieve

shetzel:
### What does this PR do?
Adds `force:mdapi:beta:retrieve` and `force:mdapi:beta:retrieve:report` commands and tests.

@W-9603705@
@W-9181617@
@W-4089551@

Note for testing:
Both of these commands can be run asynchronously by adding `-w 0` as a command parameter and should be tested.  Also, the retrieve command writes to the stash and the report command reads from it and should be tested.


src/commands/force/mdapi/beta/retrieve.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  private resolveProjectPath(): string {
+    try {
+      return SfdxProject.getInstance().getDefaultPackage().fullPath;
+    } catch (error) {
+      this.logger.debug('No SFDX project found for default package directory');
+    }
+  }
+
+  private ensureFlagPath(options: EnsureFlagOptions): string {
```
shetzel, line 235: I'm going to move this function and associated messages to `SourceCommand`

src/commands/force/mdapi/beta/retrieve/report.ts
```diff
@@ -0,0 +1,219 @@
...
+    return this.formatResult();
+  }
+
+  protected async report(): Promise<void> {
+    // Get stashed retrieve data
+    const mdRetrieveStash = Stash.get<MdRetrieveData>('MDAPI_RETRIEVE');
+
+    const retrieveId = this.resolveRetrieveId(mdRetrieveStash);
+    this.retrieveTargetDir = this.resolveOutputDir(mdRetrieveStash);
+    this.zipFileName = this.getFlag<string>('zipfilename') || mdRetrieveStash.zipfilename || 'unpackaged.zip';
```
mshanemc, line 95: let's always use `??` (it shouldn't matter for a string type, but it's a good habit)

src/commands/force/mdapi/beta/retrieve/report.ts
```diff
@@ -0,0 +1,219 @@
...
+    this.unzip = this.getFlag<boolean>('unzip') || mdRetrieveStash.unzip;
```
mshanemc, line 96: except maybe this is a `||`...are you saying that we'd want
[no unzip flag provided but unzip in stash] => unzip ?

How would that work if I have a stash left over from last week, but I actually don't want something unzipped?  That option would be gone, right?

mshanemc: Should the stash really come into play at all (for unzip and targetdir etc) if an ID is provided?

messages/md.retrieve.json
```diff
@@ -0,0 +1,48 @@
...
+    "unzip": "Extract all files from the retrieved zip file.",
+    "wait": "The number of minutes to wait for the command to complete.",
+    "apiversion": "Use to override the default, which is the latest version supported by your CLI plug-in, with the version in your package.xml file.",
+    "verbose": "Indicates that you want verbose output from the retrieve operation.",
+    "jobid": "The job ID (asyncId) of the retrieve you want to check. If not specified, the default value is the ID of the most recent metadata retrieval you ran using Salesforce CLI. You must specify a --retrievetargetdir. Use with --wait to resume waiting."
+  },
+  "InvalidFlagPath": "The %s command parameter specifies an invalid path: %s\n%s",
+  "InvalidPackageNames": "You specified [%s]. Try again and specify only one package when using --singlepackage.",
+  "MissingRetrieveId": "The jobid command parameter was not provided, neither directly nor from a previous retrieval.",
+  "checkStatus": "To check the status of this retrieve, run \"sfdx force:mdapi:beta:retrieve:report %s\""
```
mshanemc, line 47: I think it's clearer to indicate that running the report command will
1. check the status
2. but also, if the retrieve finished, store the output in your target dir, maybe unzipped if you said to

src/commands/force/mdapi/beta/retrieve/report.ts
```diff
@@ -0,0 +1,219 @@
...
+      resolvedPath = resolve(trimmedPath);
+    }
+
+    try {
+      const stats = fs.statSync(resolvedPath);
+      if (type !== 'any') {
+        const isDir = stats.isDirectory();
+        if (type === 'dir' && !isDir) {
+          const msg = 'Expected a directory but found a file';
+          throw SfdxError.create('@salesforce/plugin-source', 'md.convert', 'InvalidFlagPath', [flagName, path, msg]);
```
mshanemc, line 185: several of these are referencing md.convert messages.  If that was intentional, you'll need to load them, too.

mshanemc: I think when you wrote these, they made sense.  As they're being used more, I think it'd be easier to 
1. make flag validation functions to use with the flags' `parse` options like `dirMustExist()` or `ensureDirExists`
2. OR make some custom flags like `directoryThatExists` or `ensuredDirectory`

I understand that probably goes in the consolidation PR but these are some opinions.  I'd really love to get these into the sfdxFlags (ex: `flags.directory` would have some options to do these types of things)

src/formatters/mdapi/retrieveResultFormatter.ts
```diff
@@ -0,0 +1,138 @@
...
+  public constructor(logger: Logger, ux: UX, options: RetrieveResultFormatterOptions, result: RetrieveResult) {
+    super(logger, ux, options);
+    // zipFile can become massive and unwieldy with JSON parsing/terminal output and, isn't useful
+    delete result.response.zipFile;
+    this.options.zipFileName ??= 'unpackaged.zip';
+    const zipFilePath = join(options.retrieveTargetDir, options.zipFileName);
+    this.result = Object.assign({}, result.response, { zipFilePath });
+
+    // grab warnings
+    const warnMessages = get(result, 'response.messages', []) as RetrieveMessage | RetrieveMessage[];
```
mshanemc, line 54: are the SDR types wrong that we have to use brittle `get` and casting?

could this not just say
```ts
this.warnings = toArray(result.response.messages ?? [])
```

src/formatters/mdapi/retrieveResultFormatter.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private displayErrors(): void {
+    // an invalid packagename retrieval will end up with a message in the `errorMessage` entry
+    const errorMessage = get(this.result, 'errorMessage') as string;
+    if (errorMessage) {
+      throw new SfdxError(errorMessage);
+    }
+    const unknownMsg: RetrieveMessage[] = [{ fileName: 'unknown', problem: 'unknown' }];
+    const responseMsgs = get(this.result, 'messages', unknownMsg) as RetrieveMessage | RetrieveMessage[];
```
mshanemc, line 133: same question here about not trusting SDR typing?

If it really isn't reliable, it's worth commenting on why and creating a WI to fix.

src/stash.ts
```diff
@@ -0,0 +1,151 @@
...
+    SOURCE_DEPLOY: 'SOURCE_DEPLOY',
+  };
+
+  // singleton instance of the stash file
+  private static instance: ConfigFile<StashFile>;
+
+  private static logger: Logger;
+
+  // A map of Command.id to stash keys
+  private static keyMap = {
```
mshanemc, line 44: I know you haven't done the UT yet, but I found in mine that the UT approach typically used (extend the command), since it didn't run from the oclif executable, didn't have the id set and I was having to set it.

Think about whether it would make sense to allow the ID to be set when using the stash rather than relying on it being done for every command in every UT that uses stash.  Or maybe that's what we want.  🤷🏻

mshanemc (changes requested):
main items:
logic around using stash on `report` when there's an ID
`get` and asserting SDR types

src/commands/force/mdapi/beta/retrieve.ts
```diff
@@ -0,0 +1,304 @@
...
+    if (this.isAsync) {
+      this.ux.stopSpinner('queued');
+    } else {
+      this.ux.setSpinnerStatus(spinnerMessages.getMessage('retrieve.polling'));
+      this.retrieveResult = await this.mdapiRetrieve.pollStatus(1000, this.wait.seconds);
+      this.ux.stopSpinner();
+    }
+  }
+
+  protected resolveSuccess(): void {
```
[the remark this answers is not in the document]
shetzel: Yes.  I have a list I'm keeping but please call them out as you see them to make sure they are on my list.

src/commands/force/mdapi/beta/retrieve.ts
```diff
@@ -0,0 +1,304 @@
...
+      usernameOrConnection: this.org.getUsername(),
+      output: this.retrieveTargetDir,
+      packageOptions: this.getFlag<string[]>('packagenames'),
+      format: 'metadata',
+      singlePackage,
+      zipFileName: this.zipFileName,
+      unzip: this.unzip,
+    });
+
+    Stash.set('MDAPI_RETRIEVE', {
```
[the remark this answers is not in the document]
shetzel: Good call, I'll update.

src/commands/force/mdapi/beta/retrieve/report.ts
shetzel: It does matter though.  I want empty string to be treated as false or unset.

src/commands/force/mdapi/beta/retrieve/report.ts
shetzel: Great point.  I think if the ID is provided and it matches the stash we can do it this way, but if it doesn't match the stash then we shouldn't use any stashed values.  I'll make that change.

messages/md.retrieve.json
shetzel: I'll update, thanks.

src/commands/force/mdapi/beta/retrieve/report.ts
shetzel: This was all for the consolidation PR to follow.

I was thinking about using flag validation as I saw how this was starting to unfold.  I think that's a better, more oclif way of doing this so that's probably the approach I'll take but I'm not 100% sure since sometimes it's ok that the directory doesn't exist because (e.g.) the command will then use the default package dir.  I'll find out soon.  :)

I'm not planning on keeping the messages in the convert message bundle, but you don't have to load them explicitly when you use the message in `SfdxError.create`.  It handles it for you.

src/formatters/mdapi/retrieveResultFormatter.ts
shetzel: That line is copied from source:retrieve.  Yes, I'd love to rely on SDR types more but they're not always friendly to work with.  I can update.

src/stash.ts
shetzel: Not sure what you mean.

src/commands/force/mdapi/beta/retrieve/report.ts
mshanemc: OH!

src/commands/force/mdapi/beta/retrieve.ts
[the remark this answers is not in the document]
shetzel: TypeScript resolution didn't like that so I'm keeping it as is rather than coercing like this:
`Stash.set(Stash.KEYS.MDAPI_RETRIEVE as StashKey, { ... });`

src/formatters/mdapi/retrieveResultFormatter.ts
```diff
@@ -51,8 +51,7 @@ export class RetrieveResultFormatter extends ResultFormatter {
     this.result = Object.assign({}, result.response, { zipFilePath });
```
mshanemc, line 51: if `result` is not guaranteed, you'd need a nullco here, too.  Or maybe you don't need it in the `this.warnings`.

src/commands/force/mdapi/beta/retrieve.ts
```diff
@@ -0,0 +1,304 @@
...
+      sourcepath: this.sourceDir ? [this.sourceDir] : undefined,
+      manifest: manifest && {
+        manifestPath: manifest,
+        directoryPaths: [],
+      },
+    });
+
+    await this.lifecycle.emit('preretrieve', { packageXmlPath: manifest });
+
+    this.ux.setSpinnerStatus(spinnerMessages.getMessage('retrieve.sendingRequest', [this.componentSet.apiVersion]));
```
mshanemc, line 149: for consolidation, maybe this belongs on sourceCommand: 
https://github.com/salesforcecli/plugin-source/blob/5f0ad449a6880d761a233c3191e3759a6c0b72cc/src/deployCommand.ts#L212

mshanemc (approved):
approved pending UT

WillieRuemmele (approved):
🚀 

pending UTs 😄

test/nuts/mdapi.nut.ts
```diff
@@ -211,8 +213,173 @@ describe('mdapi NUTs', () => {
...
+
+        // Async report, from stash
+        let reportCmd = 'force:mdapi:beta:retrieve:report -w 0 --json';
+        const rv2 = execCmd<RetrieveCommandAsyncResult>(reportCmd, { ensureExitCode: 0 });
+        const result2 = rv2.jsonOutput.result;
+        expect(result2).to.have.property('done', false);
+        expect(result2).to.have.property('id', result1.id);
+        // To prevent flapping we expect 1 of 2 likely states.  All depends
+        // on how responsive the message queue is.
+        expect(result2.state).to.be.oneOf(['Queued', 'InProgress']);
```
jshackell-sfdc, line 314:
```suggestion
        expect(result2.state).to.be.oneOf(['Queued', 'In Progress']);
```

test/nuts/mdapi.nut.ts
```diff
@@ -211,8 +213,173 @@ describe('mdapi NUTs', () => {
...
+        expect(result2.status).to.be.oneOf(['Queued', 'InProgress']);
```
jshackell-sfdc, line 315:
```suggestion
        expect(result2.status).to.be.oneOf(['Queued', 'In Progress']);
```

messages/md.retrieve.json
```diff
@@ -0,0 +1,48 @@
+{
+  "retrieveCmd": {
+    "description": "retrieve metadata from an org using Metadata API\nUses Metadata API to retrieve a .zip of XML files that represent metadata from the targeted org. The default target username is the admin user for the default scratch org. You can retrieve and deploy up to 10,000 files or 400 MB (39 MB compressed) at one time.",
+    "examples": [
```
jshackell-sfdc, line 4: Can we add the following brief explanations before each example?  This mirrors what Shane did with mdapi:deploy (please correct if my desc is wrong):

"Retrieve metadata using specified target directory:",
"Specify a manifest file:",
"Give a custom name to the retrieved .zip file and automatically extract its contents:",

src/commands/force/mdapi/beta/retrieve.ts
```diff
@@ -0,0 +1,304 @@
...
+      longDescription: messages.getMessage('flagsLong.packagenames'),
+      exclusive: ['sourcedir', 'unpackaged'],
+    }),
+    singlepackage: flags.boolean({
+      char: 's',
+      description: messages.getMessage('flags.singlepackage'),
+      longDescription: messages.getMessage('flagsLong.singlepackage'),
+    }),
+    zipfilename: flags.string({
+      char: 'n',
```
jshackell-sfdc, line 72:
```suggestion
      char: 'f',
```

jshackell-sfdc: We use "n" for names, but "f" for filenames.  At least, according to this: https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_plugins.meta/sfdx_cli_plugins/cli_plugins_style_parameter_names.htm

messages/deploy.json
```diff
@@ -75,7 +75,6 @@
   },
   "checkOnlySuccess": "Successfully validated the deployment. %s components deployed and %s tests run.\nUse the --verbose parameter to see detailed output.",
   "MissingDeployId": "No deploy ID was provided or found in deploy history",
```
jshackell-sfdc, line 77:
```suggestion
  "MissingDeployId": "No deploy ID was provided or found in deploy history.",
```

messages/deploy.json
```diff
@@ -75,7 +75,6 @@
...
-  "InvalidStashFile": "Invalid deployment stash file encountered. File has been renamed to: %s.  Please specify a deploy ID using the command parameter.",
   "deployCanceled": "The deployment has been canceled by %s",
```
jshackell-sfdc, line 78:
```suggestion
  "deployCanceled": "The deployment has been canceled by %s.",
```

messages/stash.json
```diff
@@ -0,0 +1,4 @@
+{
+  "InvalidStashKey": "The stash key provided is invalid.  Found: [%s]",
```
jshackell-sfdc, line 2:
```suggestion
  "InvalidStashKey": "The stash key provided is invalid.  Found: [%s].",
```

src/commands/force/mdapi/beta/retrieve/report.ts
```diff
@@ -0,0 +1,221 @@
...
+      description: messages.getMessage('flags.retrievetargetdir'),
+      longDescription: messages.getMessage('flagsLong.retrievetargetdir'),
+    }),
+    jobid: flags.id({
+      char: 'i',
+      description: messages.getMessage('flags.jobid'),
+      longDescription: messages.getMessage('flagsLong.jobid'),
+    }),
+    zipfilename: flags.string({
+      char: 'n',
```
jshackell-sfdc, line 53:
```suggestion
      char: 'f',
```

WillieRuemmele:
`beta:retrieve`:
the "all" directory referenced was generated with `force:source:convert -p force-app` in dreamhouse
the "objects" directory is generated from `convert -p force-app/main/default/objects`
"manifest" is generated from `manifest:create -p force-app`


✅ : `-a behavior`: now with warnings :)
✅ :`-r` behavior consistent
✅ : `-k` behavior consistent
✅ : `-d` behavior consistent
✅ : `--unzip`: new behavior 😄 
✅ : `-p` behavior consistent
✅  : `--zipfilename`: works
❓/ ✅ : `--singlepackage` behavior is consistent, even when working with MPD
✅ : `-w` works, enqueues async request, or will wait for request to finish
✅ : `--json` output
✅ : `--verbose` -> more output

`retrieve:report`:
✅ : Using Steve's script
✅ : `-w` works
✅ : uses stash correctly
❌ : `--json` missing the 
```    "zipFile": {
      "$": {
        "xsi:nil": "true"
      }
    },
``` 
property. I ran `sfdx force:mdapi:retrieve -k classes.xml -r allclasses -w 0 ; sfdx force:mdapi:retrieve:report -w 0 --json | pbcopy ` vs `sfdx force:mdapi:beta:retrieve -k classes.xml -r allclasses -w 0 ; sfdx force:mdapi:retrieve:beta:report -w 0 --json | pbcopy `... I'm not sure it _really_ matters 🤷‍♂️

messages/md.retrieve.json
```diff
@@ -0,0 +1,48 @@
...
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir",
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir -k package.xml",
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir --unzip --zipfilename apexClasses.zip"
+    ]
+  },
+  "reportCmd": {
+    "description": "check the status of a metadata retrieval\nSpecify the job ID and a target directory for the retrieve you want to check. You can also specify a wait time (minutes) to check for updates to the retrieve status. If the retrieve was successful, the resulting zip file will be saved to the location passed in with the retrieve target parameter.",
+    "examples": [
```
jshackell-sfdc, line 12: Add example descriptions before each example (please fix if I got it wrong):

"Check the status of the most recent deploy:",
"Wait for 15 minutes, and if the request is complete, retrieve the files into the specified directory:",
"Check status, and if the request is complete, give the zip a custom name and automatically unzip it:"

messages/md.retrieve.json
```diff
@@ -0,0 +1,48 @@
...
+  "retrieveCmd": {
+    "description": "retrieve metadata from an org using Metadata API\nUses Metadata API to retrieve a .zip of XML files that represent metadata from the targeted org. The default target username is the admin user for the default scratch org. You can retrieve and deploy up to 10,000 files or 400 MB (39 MB compressed) at one time.",
+    "examples": [
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir",
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir -k package.xml",
+      "$ sfdx force:mdapi:beta:retrieve -r path/to/retrieve/dir --unzip --zipfilename apexClasses.zip"
+    ]
+  },
+  "reportCmd": {
+    "description": "check the status of a metadata retrieval\nSpecify the job ID and a target directory for the retrieve you want to check. You can also specify a wait time (minutes) to check for updates to the retrieve status. If the retrieve was successful, the resulting zip file will be saved to the location passed in with the retrieve target parameter.",
```
jshackell-sfdc, line 11:
```suggestion
    "description": "check the status of a metadata retrieval\nSpecify the job ID and a target directory for the retrieve you want to check; default is the most recent deploy. If the retrieve was successful, the resulting zip file is saved to the location specified by the -r parameter.",
```

messages/md.retrieve.json
```diff
@@ -0,0 +1,48 @@
...
+    "unzip": "extract all files from the retrieved zip file",
+    "wait": "wait time for command to finish in minutes",
+    "apiversion": "target API version for the retrieve",
+    "verbose": "verbose output of retrieve result",
+    "jobid": "job ID of the retrieve you want to check; defaults to your most recent CLI retrieval if not specified"
+  },
+  "flagsLong": {
+    "retrievetargetdir": "The root of the directory structure where the retrieved .zip or metadata files are put.",
+    "unpackaged": "The complete path for the manifest file that specifies the components to retrieve.",
+    "sourcedir": "The source directory to use instead of the default package directory specified in sfdx-project.json",
```
jshackell-sfdc, line 34:
```suggestion
    "sourcedir": "The source directory to use instead of the default package directory specified in sfdx-project.json.",
```

PR #402, opened 2022-01-24
Sm/deploy-concise

mshanemc:
### What does this PR do?
support consise json for mdapi:deploy|report

@W-4416089@


WillieRuemmele (approved):

WillieRuemmele:
QA:
✅ : `--concise` has no effect on stdout. It's already pretty concise with just the progress bar and deploy ID. I could see how this flag would've been more useful before we used the progress bar
✅ : removes `componentSuccess` from JSON output
❌ : `SFDX_USE_PROGRESS_BAR=false sfdx force:mdapi:beta:deploy -d mdapi/ -w 4`, should disable a lot of the output. I think this is what the bug is referring to in reducing the amount of output (before we used the progress bar). Maybe only show the initial status, and the success status.

PR #419, opened 2022-02-03
Sm/qa pr 418

mshanemc:
### What does this PR do?
open a vf page by name

@W-10514651@
https://github.com/forcedotcom/cli/issues/1386


mshanemc:
QA Notes:

✅ opens a page using the .page file
👎🏻  sends user to odd place with message `The name can only contain underscores and alphanumeric characters. It must begin with a letter and be unique, and must not include spaces, end with an underscore, or contain two consecutive underscores.` when opening the meta-xml
👎🏻 opening a non-existent page (ex: mistyping a filename) sends user to a login page

mshanemc:
✅ opens vf meta files
✅ throws when file doesn't exist

src/commands/force/source/open.ts
```diff
@@ -98,18 +101,27 @@ export class Open extends SourceCommand {
...
+              '.flexipage-meta.xml'
+            )}'`,
+            { tooling: true }
+          );
+        return `/visualEditor/appBuilder.app?pageId=${flexipage.Id}`;
+      } else if (pageType === 'ApexPage') {
+        return `/apex/${path
+          .basename(this.flags.sourcefile as string)
+          .replace('.page-meta.xml', '')
+          .replace('.page', '')}`;
```
peternhale, line 121:
``` suggestion
          .replace(/\.page-meta\.xml|\.page/, '')}`;
```

mshanemc: 🤷🏻‍♂️   that's a close one.  I think I'll leave as-is for readability, since perf is negligible here.

PR #423, opened 2022-02-12
feat: create manifest from targetUsername (ComponentSet.fromConnection)

jayree:
### What does this PR do?

proposal to use the new ComponentSet.fromConnection function.

### DIFF
--- a/messages/create.json
+++ b/messages/create.json
@@ -7,8 +7,8 @@
     "$ sfdx force:source:manifest:create --fromorg test@myorg.com --includepackages unlocked"
   ],
   "flags": {
-    "includepackages": "include metadata from 'managed' or 'all' package types",
-    "fromorg": "build a manifest from all source in the target org",
+    "includepackages": "comma-separated list of package types whose metadata is included in the manifest; by default, metadata in packages is ignored",
+    "fromorg": "username or alias of the org that contains the metadata components from which to build a manifest",
     "manifesttype": "type of manifest to create; the type determines the name of the created file",
     "manifestname": "name of a custom manifest file to create",
     "outputdir": "directory to save the created manifest",
--- a/messages/create.json
+++ b/messages/create.json
@@ -3,7 +3,8 @@
   "examples": [
     "$ sfdx force:source:manifest:create -m ApexClass",
     "$ sfdx force:source:manifest:create -m ApexClass:MyApexClass --manifesttype destroy",
-    "$ sfdx force:source:manifest:create --sourcepath force-app --manifestname myNewManifest"
+    "$ sfdx force:source:manifest:create --sourcepath force-app --manifestname myNewManifest",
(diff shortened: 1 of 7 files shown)


src/commands/force/source/manifest/create.ts
```diff
@@ -22,16 +22,22 @@ const manifestTypes: Record<string, string> = {
...
   path: string;
 }
 
-const xorFlags = ['metadata', 'sourcepath'];
+const xorFlags = ['metadata', 'sourcepath', 'targetusername'];
 export class create extends SourceCommand {
   public static readonly description = messages.getMessage('description');
   public static readonly examples = messages.getMessage('examples').split(os.EOL);
   public static readonly requiresProject = true;
+  public static readonly supportsUsername = true;
```
shetzel, line 40: This will cause issues with default usernames when this command runs, and we'd like to change the flag description to be specific for this command so we're suggesting to add another string type flag named, `fromorg` with no short char rather than use `supportsUsername`

jayree:
@shetzel I've added a new commit which contains the requested changes. Please review again. Thank you!

messages/create.json
```diff
@@ -6,6 +6,8 @@
     "$ sfdx force:source:manifest:create --sourcepath force-app --manifestname myNewManifest"
   ],
   "flags": {
+    "includepackages": "include metadata from 'managed' or 'all' package types",
+    "fromorg": "build a manifest from all source in the target org",
```
jshackell-sfdc, line 10:
```suggestion
    "fromorg": "scratch org or sandbox that contains the metadata components from which to build a manifest",
```

messages/create.json
```diff
@@ -6,6 +6,8 @@
     "$ sfdx force:source:manifest:create --sourcepath force-app --manifestname myNewManifest"
   ],
   "flags": {
+    "includepackages": "include metadata from 'managed' or 'all' package types",
```
jshackell-sfdc, line 9:
```suggestion
    "includepackages": "comma-separated list of package types whose metadata is included in the manifest; by default, metadata in packages is ignored",
```

messages/create.json
```diff
@@ -6,6 +6,8 @@
     "$ sfdx force:source:manifest:create --sourcepath force-app --manifestname myNewManifest"
```
jshackell-sfdc, line 6: Can we add an example of generating a manifest from an org? Something like:

$ sfdx force:source:manifest:create --fromorg test@myorg.com --includepackages managed

messages/create.json
jshackell-sfdc: @shetzel , I included "alias" in the --fromorg description.  If it doesn't work with aliases, please remove.

jshackell-sfdc (changes requested):
See my comments about the --help messages.

messages/create.json
jayree: I've added the example.

shetzel:
Merged with https://github.com/salesforcecli/plugin-source/pull/429

PR #426, opened 2022-02-18
feat: add --retrievetargetdir to source:retrieve

amtrack:
### What does this PR do?

Using this flag, it should be possible to explicitly retrieve and place source files to a given package directory (or non package directory).

I'm opening this PR to get some early feedback from you.

> :question: Do you think this could be useful for `plugin-source` or should I rather put this in an unofficial sfdx plugin?

use cases

1. "poor mans code completion" for ApexClasses from Unlocked Packages (works with VS Code)

   ```console
   sfdx force:source:retrieve -r force-app/.tmp -m "$(sfdx-plugin-package-xml force:mdapi:listallmetadata -m ApexClass -- 
   unlocked --output name-csv)"
   ```

1. retrieving and updating "unpackaged" source files in a separate directory that should not be pushed to a scratch org (no entry in `sfdx-project.json`)

   ```console
   sfdx force:source:retrieve -r ./unpackaged -m "Settings"
   ```

references

- https://salesforce.stackexchange.com/questions/308471/can-you-set-the-target-directory-for-forcesourceretrieve-operations
- https://github.com/msrivastav13/mo-dx-plugin#sfdx-retrievedxsource

TODO

- [ ] omit directory structure (`main/default`)?

   Currently If I run

   ```console
   sfdx force:source:retrieve -m StandardValueSet:TaskStatus -r ./unpackaged
   ```

   the source file will currently be placed in (red) but I would expect it to be placed in (green):

   ```diff
   - unpackaged/main/default/standardValueSets/TaskStatus.standardValueSet-meta.xml
   + unpackaged/standardValueSets/TaskStatus.standardValueSet-meta.xml
   ```

   Is it possible to omit `main/default` in that case?

- [x] ~~improve output of command when source files already exist in package directories~~

   ~~Currently, if the source file already exists in a package directory, the output of the command using `--retrievetargetdir` indicates that no Metadata has been retrieved.~~
Apparently this was only the case when using `-p` which is now disallowed in combination with `-r`.


mshanemc:
Have you tried `-r` with `-p` to see what happens?  If it's crazy, you could prevent those 2 flags from being used together (forcing users into using -m or a manifest)

I don't think there's a way to `omit directory structure (main/default)`.  That logic lives over in SDR...you can start here to follow the behavior: 
https://github.com/forcedotcom/source-deploy-retrieve/blob/407af6754994b9b8b0484ee7a8dcb380ada2c554/src/resolve/sourceComponent.ts#L166

> Currently, if the source file already exists in a package directory, the output of the command using --retrievetargetdir indicates that no Metadata has been retrieved.

That's another SDR behavior.  It's going to report on the files that it **wrote** not what it actually retrieved.

messages/retrieve.json
```diff
@@ -24,6 +26,7 @@
     "forceoverwrite": "ignore conflict warnings and overwrite changes to the project"
   },
   "flagsLong": {
+    "retrievetargetdir": "The root of the directory structure where the source files are put.",
```
jshackell-sfdc, line 29:
```suggestion
    "retrievetargetdir": "The root of the directory structure into which the source files are retrieved.",
```

amtrack:
@mshanemc Thank you for your very useful feedback!

> That's another SDR behavior. It's going to report on the files that it wrote not what it actually retrieved.

Apparently this was only the case when using `-p` which is now disallowed in combination with `-r`.
So I think this is not an issue anymore.

The only thing that still bothers me is the `main/default` in the path. I'll need to look into the SDR...

amtrack:
Hi @peternhale, thanks for taking this over!
It is kind of a showstopper for me. 😄 
Personally I'll probably be able to deal with `main/default` subdirs by running a `mv` command afterwards or s.th. like that.

But In my opinion I have never had a benefit of the `main/default` subdir and wouldn't like to see it spreading more than necessary.
And if people will have used it for a while it might get more difficult to change that behaviour in the future.

A real solution would be to make SDR use the `main/default` (`DEFAULT_PACKAGE_ROOT_SFDX`) only in the context of a default package directory, but I don't know how.

A trade-off could be a more detailed flag description (s.th. like "Currently this will create a main/default subdirectory in the target directory. This behaviour might change in the future.").

peternhale:
@amtrack We are actively discussing what could be done with `main/default`. You may see the feature show up w/o a decision though, at least in this iteration. We appreciate your contribution.

PR #429, opened 2022-02-23
Create manifest from targetusername

shetzel:
### What does this PR do?
Adds flags to the existing `force:source:manifest:create` command to create a manifest from metadata in an org, optionally including packaged metadata.

@W-10731076@

All credit goes to @jayree and his work in the SDR library and this PR: https://github.com/salesforcecli/plugin-source/pull/423


shetzel:
Testing:
✅ create from org alias
✅ create from org username
✅ create from org alias and `--includepackages foo` properly throws error
✅ create from org alias and `--includepackages unlocked`
✅ create from org alias and `--includepackages managed,unlocked`
✅ create from org alias and `--includepackages managed,unlocked,foo` properly throws error
✅ create from org alias and `--includepackages "managed, unlocked"`
All NUTs pass
All unit tests pass

PR #471, opened 2022-04-25
fix: pull non-default directory in default project

mshanemc:
### What does this PR do?
use new STL method for generating a componentSet that mixes SourceComponent (where the exist) with metadataMember (where they are remote only) to ensure that the retrieved components don't duplicate to default dir

@W-11047555@
https://github.com/forcedotcom/cli/issues/1485

### DIFF
--- a/src/commands/force/source/pull.ts
+++ b/src/commands/force/source/pull.ts
@@ -8,14 +8,8 @@
 import { flags, FlagsConfig } from '@salesforce/command';
 import { Duration } from '@salesforce/kit';
 import { Messages } from '@salesforce/core';
-import {
-  ComponentSet,
-  FileResponse,
-  RequestStatus,
-  RetrieveResult,
-  SourceComponent,
-} from '@salesforce/source-deploy-retrieve';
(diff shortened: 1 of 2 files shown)


src/commands/force/source/pull.ts
```diff
@@ -85,22 +79,9 @@ export default class Pull extends SourceCommand {
...
-      if (component.type && component.name) {
-        componentSet.add({
-          type: component.type,
-          fullName: component.name,
-        });
-      }
-    });
+    const componentSet = await this.tracking.remoteNonDeletesAsComponentSet();
 
+    // if it is't local, add it as a
```
shetzel, line 84: Should this just be removed?

shetzel (approved):
I can merge as is, but that comment either needs updating or removal.  There should be a NUT added to cover this usecase as well, but that can go in another PR in the interest of getting this in for RC.

cromwellryan:
@shetzel , I see a few questions about NUTS and comments that need removed. I don't see a response. Did you intend to move this through? Can either you or @mshanemc clear up the confusion if someone on the team or community looks into this PR for clarity.

PR #538, opened 2022-07-14
fix: try/finally on pollStatus to handle throws from timeout

mshanemc:
### What does this PR do?
try/finally on pollStatus for both `mdapi:deploy:report` and `source:deploy:report`
after the deploy times out, it uses the `this.report` one more time to get a good in-progress DeployResult

https://github.com/forcedotcom/cli/issues/1612
[@W-11431588@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000011TwIzYAK/view)

### QA notes

I used [EDA](https://github.com/mshanemc/EDA) to get a really slow deploy like the customer has

```
# eda is weird with its scratch orgs and namespace, so
sfdx force:org:create -d 1 -f orgs/dev.json --nonamespace -s
# convert source for mdapi deploy
sfdx force:source:convert -d force-app --outputdir=mdapiOut
``` 

steps based on the customer's logic.  Have these ready to copy/paste quickly
```
../../plugin-source/bin/dev force:mdapi:deploy --deploydir mdapiOut --ignorewarnings -w 0 --testlevel RunLocalTests --json
# but use the job id from your deploy
../../plugin-source/bin/dev force:mdapi:deploy:report -i 0Af1k00001PyBXhCAN --wait 1 --json
# verifying fix for source:deploy
../../plugin-source/bin/dev force:source:deploy:report -i 0Af1k00001PyBXhCAN --wait 1 --json
```
Screen Shot 2022-07-14 at 2 35 37 PM

### DIFF
--- a/src/commands/force/mdapi/deploy/report.ts
+++ b/src/commands/force/mdapi/deploy/report.ts
@@ -82,11 +82,9 @@ export class Report extends DeployCommand {
     );
 
     if (this.isAsync) {
-      this.deployResult = await this.report(this.getFlag<string>('jobid'));
+      this.deployResult = await this.report(deployId);
       return;
     }
-
-    this.displayDeployId(deployId);
 
     const deploy = this.createDeploy(deployId);
     if (!this.isJsonOutput()) {
@@ -95,7 +93,12 @@ export class Report extends DeployCommand {
         : new DeployProgressStatusFormatter(this.logger, this.ux);
       progressFormatter.progress(deploy);
     }
-    this.deployResult = await deploy.pollStatus({ frequency: Duration.milliseconds(500), timeout: waitDuration });
+
+    try {
+      await deploy.pollStatus({ frequency: Duration.milliseconds(500), timeout: waitDuration });
+    } finally {
+      this.deployResult = await this.report(deployId);
+    }
   }
 
   // this is different from the source:report uses report error codes (unfortunately)
--- a/src/commands/force/source/deploy/report.ts
+++ b/src/commands/force/source/deploy/report.ts
@@ -93,8 +93,11 @@ export class Report extends DeployCommand {
         : new DeployProgressStatusFormatter(this.logger, this.ux);
       progressFormatter.progress(deploy);
     }
-    await deploy.pollStatus({ timeout: waitDuration });
-    this.deployResult = await this.report(deployId);
+    try {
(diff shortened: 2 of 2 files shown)


src/commands/force/mdapi/deploy/report.ts
```diff
@@ -82,20 +82,23 @@ export class Report extends DeployCommand {
...
         ? new DeployProgressBarFormatter(this.logger, this.ux)
         : new DeployProgressStatusFormatter(this.logger, this.ux);
       progressFormatter.progress(deploy);
     }
-    this.deployResult = await deploy.pollStatus({ frequency: Duration.milliseconds(500), timeout: waitDuration });
+
+    try {
+      await deploy.pollStatus({ frequency: Duration.milliseconds(500), timeout: waitDuration });
+    } finally {
+      this.deployResult = await this.report(deployId);
```
shetzel, line 100: Couldn't we run into a scenario where `pollStatus()` is successful but then we run `this.report()` and get a connection error?

mshanemc: Good point.  `source:deploy` already was already open to that potential issue.

Would you rather try to set `this.deployResult` from pollStatus, and then finally would call this.report only if needed?

The problem is this.report has all kinds of side effects beyond checking status (handling undefined id from the stash, printing to the terminal, passing the componentSet part to SDR/DeployResult's constructor, etc).

shetzel (approved):

src/commands/force/mdapi/deploy/report.ts
```diff
@@ -95,9 +95,15 @@ export class Report extends DeployCommand {
...
-    } finally {
-      this.deployResult = await this.report(deployId);
+      this.deployResult = await deploy.pollStatus({ frequency: Duration.milliseconds(500), timeout: waitDuration });
+    } catch (error) {
+      const err = error as Error;
+      if (err.message.includes('The client has timed out')) {
+        this.logger.debug('mdapi:deploy:report polling timed out. Requesting status...');
+        this.deployResult = await this.report(deployId);
+      } else {
+        throw err;
```
mshanemc, line 105:
```suggestion
      if (error instanceof Error && error.message.includes('The client has timed out')) {
        this.logger.debug('mdapi:deploy:report polling timed out. Requesting status...');
        this.deployResult = await this.report(deployId);
      } else {
        throw error;
```

PR #539, opened 2022-07-21
fix: testing mocha retries

iowillhoit:


iowillhoit:
Closing since mocha retries do not re-run any hooks (before, after, etc). Rerunning tests are unlikely to pass if the setup is not re-running.

PR #553, opened 2022-08-11
fix: exclude sourcepath from component set builder

shetzel:
### What does this PR do?
Exclude passing a sourcepath to the ComponentSetBuilder when the `unpackaged` flag is set.  The `unpackaged` flag is the path to the manifest and the command should only use the manifest, not consider any local source path.

@W-11588116@

### DIFF
--- a/src/commands/force/mdapi/retrieve.ts
+++ b/src/commands/force/mdapi/retrieve.ts
@@ -110,7 +110,7 @@ export class Retrieve extends SourceCommand {
 
   protected async retrieve(): Promise<void> {
     const packagenames = this.getFlag<string[]>('packagenames');
-    if (!packagenames) {
+    if (!packagenames && !this.flags.unpackaged) {
       this.sourceDir = this.resolveRootDir(this.getFlag<string>('sourcedir'));
     }
     this.retrieveTargetDir = this.resolveOutputDir(this.getFlag<string>('retrievetargetdir'));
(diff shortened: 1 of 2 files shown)


src/commands/force/mdapi/retrieve.ts
```diff
@@ -110,7 +110,7 @@ export class Retrieve extends SourceCommand {
 
   protected async retrieve(): Promise<void> {
     const packagenames = this.getFlag<string[]>('packagenames');
-    if (!packagenames) {
+    if (!packagenames && !this.flags.unpackaged) {
```
mshanemc, line 113: yep, that should solve it.

mshanemc (approved):

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes:
---
```
 ➜  sfdx force:mdapi:retrieve -r force-app -k package.xml                                        
ERROR running force:mdapi:retrieve:  /Users/william.ruemmele/projects/scratches/dreamhouse-lwc/force-app/main/default/staticresources/sample_data_properties.resource-meta.xml: Expected source files for type 'StaticResource'
Retrieving metadata from test-anyzpqmkqve1@example.com... done
➜  dreamhouse-lwc git:(main) ✗  hub:(GLOBAL - DevHub) scratch:(test-anyzpqmkqve1@example.com)
 ➜  ../../oss/plugin-source/bin/run force:mdapi:retrieve -r force-app -k package.xml                                        
Retrieve ID: 09S0t000002ySmLEAU
Retrieving metadata from test-anyzpqmkqve1@example.com... done
Wrote retrieve zip to /Users/william.ruemmele/projects/scratches/dreamhouse-lwc/force-app/unpackaged.zip
```
previously would parse the local FS to build a comp. set,
now just requests what's in the manifest

PR #629, opened 2022-10-28
fix: add hook and doctor diagnostic tests

shetzel:
### What does this PR do?
Adds a hook listener for sfdx/sf CLI doctor command executions to run diagnostic tests specific to this plugin.
There is now 1 diagnostic test that inspects the user environment for `apiVersion`, `sourceApiVersion`, and the max apiVersion of the default target DevHub org and the default target org.  The test adds this data to the doctor's diagnostic report, adds test results comparing the various apiVersions, and adds suggestions based on the test results.
There are also unit tests for the diagnostic tests to ensure their behavior.

@W-11949474@


peternhale (approved):

messages/diagnostics.json
```diff
@@ -0,0 +1,7 @@
+{
+  "apiVersionMismatch": "The sourceApiVersion in sfdx-project.json does not match the apiVersion. This means deploying and retrieving source will use sourceApiVersion. This may or may not be desired.",
```
jshackell-sfdc, line 2:
```suggestion
  "apiVersionMismatch": "The sourceApiVersion in sfdx-project.json doesn\'t match the apiVersion. The commands that deploy and retrieve source use the sourceApiVersion in this case. The version mismatch isn't a problem, as long as it\'s the behavior you actually want. ",
```

messages/diagnostics.json
```diff
@@ -0,0 +1,7 @@
...
+  "apiVersionUnset": "Both sourceApiVersion and apiVersion are not defined. This means deploying and retrieving source will use the max apiVersion of the target org. This may or may not be desired.",
```
jshackell-sfdc, line 3:
```suggestion
  "apiVersionUnset": "Neither sourceApiVersion nor apiVersion are defined. The commands that deploy and retrieve source use the max apiVersion of the target org in this case. The issue isn't a problem, as long as it's the behavior you actually want.",
```

messages/diagnostics.json
```diff
@@ -0,0 +1,7 @@
...
+  "maxApiVersionMismatch": "The max apiVersion for the default DevHub org does not match the max apiVersion for the default target org. This means the default target orgs are running different versions and without explicitly setting an apiVersion there may be deploy/retrieve issues.",
```
jshackell-sfdc, line 4:
```suggestion
  "maxApiVersionMismatch": "The max apiVersion of the default DevHub org doesn't match the max apiVersion of the default target org. This mismatch means that the default target orgs are running different API versions. Be sure you explicitly set the apiVersion when you deploy or retrieve source, or you will likely run into problems.",
```

jshackell-sfdc: I changed the suggestion into a positive (i.e. "set the apiVersion") rather than a negative.  Okay?

messages/diagnostics.json
```diff
@@ -0,0 +1,7 @@
...
+  "sourceApiVersionMaxMismatch": "The sourceApiVersion in sfdx-project.json does not match the max apiVersion for the default target org. This means you are not using the latest features available from API version %s. This may or may not be desired.",
```
jshackell-sfdc, line 5:
```suggestion
  "sourceApiVersionMaxMismatch": "The sourceApiVersion in sfdx-project.json doesn't match the max apiVersion of the default target org. As a result, you're not using the latest features available from API version %s. The version mismatch isn't a problem, as long as it's the behavior you actually want.",
```

messages/diagnostics.json
```diff
@@ -0,0 +1,7 @@
...
+  "apiVersionMaxMismatch": "The apiVersion in the configuration does not match the max apiVersion for the default target org. This means you are not using the latest features available from API version %s. This may or may not be desired."
```
jshackell-sfdc, line 6:
```suggestion
  "apiVersionMaxMismatch": "The apiVersion doesn't match the max apiVersion of the default target org. As a result, you're not using the latest features available in API version %s. The version mismatch isn't a problem, as long as it's the behavior you actually want."
```

PR #755, opened 2023-03-08
Wr/deprecate and delete

WillieRuemmele:
### What does this PR do?
deprecates commands that are available in PDR

removes commands moved to PDR

@W-12566134@


mshanemc (changes requested):
I'd like to use these tests to test Mike's patterns and see what doesn't work.  I'm goign to spend some time on those next week.

mshanemc:
> I'd like to use these tests to test Mike's patterns and see what doesn't work. I'm goign to spend some time on those next week.

welp, I got 1 test done.

PR #815, opened 2023-04-14
fix: pass --soapdeploy flag to method, add org-metadata

WillieRuemmele:
### What does this PR do?
passes `--soapdeploy` flag to method that requires it
adds support for `org-metadata-rest-deploy` config where it was using `SfdxConfigAggergator`

@W-13036560@
https://github.com/forcedotcom/cli/issues/2062


src/deployCommand.ts
```diff
@@ -121,7 +121,9 @@ export abstract class DeployCommand extends SourceCommand {
       return false;
     }
 
-    const restDeployConfig = this.configAggregator.getInfo(SfdxPropertyKeys.REST_DEPLOY).value;
+    const restDeployConfig =
```
WillieRuemmele, line 124: if you run `sf force:source:deploy` the `configAggregator` will be a `ConfigAggregator` not a `SfdxConfigAggregator` which won't know about the `sf` config values

mshanemc (commented):
I have no objection to this fix going out, but if we get ConfigAggregator fixed, we don't need it.

mshanemc (approved):

mshanemc:
QA:

## source:deploy

✅  no flags, no config => soap
❌  sfdx config set org-metadata-rest-deploy=true, no flags => soap
❌  sfdx config set restDeploy=true, no flags => soap

## mdapi:deploy after convert
❌ both new and old config set to true, globally => soap

I stopped here because I can't get anything to do a rest deploy.

PR #821, opened 2023-04-18
fix: remove -o conflict, add short char UTs, add short chars to NUTs

WillieRuemmele:
### What does this PR do?
removes `-o` char duplication on `source/mdapi:deploy`
adds NUTs that uses short options
adds UTs that use short options

@W-13056553@

### DIFF
--- a/src/commands/force/mdapi/deploy.ts
+++ b/src/commands/force/mdapi/deploy.ts
@@ -12,11 +12,16 @@ import {
   Flags,
   loglevel,
   orgApiVersionFlagWithDeprecations,
-  requiredOrgFlagWithDeprecations,
   Ux,
 } from '@salesforce/sf-plugins-core';
 import { Interfaces } from '@oclif/core';
-import { DeployCommand, getCoverageFormattersOptions, reportsFormatters, TestLevel } from '../../../deployCommand';
+import {
+  DeployCommand,
+  getCoverageFormattersOptions,
+  reportsFormatters,
+  targetUsernameFlag,
+  TestLevel,
+} from '../../../deployCommand';
 import { DeployCommandAsyncResult } from '../../../formatters/source/deployAsyncResultFormatter';
 import { MdDeployResult, MdDeployResultFormatter } from '../../../formatters/mdapi/mdDeployResultFormatter';
 import { ProgressFormatter } from '../../../formatters/progressFormatter';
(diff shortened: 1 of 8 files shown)


src/deployCommand.ts
```diff
@@ -259,3 +271,26 @@ export const getCoverageFormattersOptions = (formatters: string[] = []): Coverag
...
+export const targetUsernameFlag = Flags.string({
+  required: true,
+  char: 'u',
+  deprecateAliases: true,
+  // DO NOT alias to 'o', it will conflict with '--ignoreerrors'
+  aliases: ['targetusername', 'u'],
+  summary: messages.getMessage('flags.targetusername.summary'),
+  parse: async (input: string | undefined) => resolveUsername(input),
+  default: async () => resolveUsername(),
+  defaultHelp: async () => resolveUsername(),
```
shetzel, line 284: Is this correct for defaultHelp?

src/commands/force/mdapi/deploy.ts
```diff
@@ -75,6 +80,7 @@ export class Deploy extends DeployCommand {
       summary: messages.getMessage('flags.runTests.summary'),
     }),
     ignoreerrors: Flags.boolean({
+      // break this
```
shetzel, line 83: What does this comment mean?

src/deployCommand.ts
WillieRuemmele: This is similar to what we did in plugin-org

https://github.com/salesforcecli/plugin-org/pull/646/files#diff-2cb023071675d0d3c17b86b021f81fd33622048a9eab2fb3d6623e52375f2690R38

src/commands/force/mdapi/deploy.ts
WillieRuemmele: leftover - will remove

shetzel (approved):

src/deployCommand.ts
peternhale: This form can leak customer data into the manifest.

src/deployCommand.ts
```diff
@@ -259,3 +271,26 @@ export const getCoverageFormattersOptions = (formatters: string[] = []): Coverag
     reportOptions,
   };
 };
+
+export const targetUsernameFlag = Flags.string({
```
peternhale, line 275: Having target-org with short char other than 'o' break sf guidelines. I ran into short char collision in plugin-user and to avoid the collision we ended up retaining the old commands that used the `-o` and created new ones that allows style adherence.

peternhale (changes requested):

src/deployCommand.ts
WillieRuemmele: @peternhale ^

peternhale (approved):

PR #828, opened 2023-04-27
Sh/push json regression

shetzel:
### What does this PR do?
Removes the `catch` function from push.  Adds the `context` property to the SfError that will now rely on `SfCommand.catch` to format an error that does not break backward compatibility.  Adds a few more tests around this.

@W-13100374@

### DIFF
--- a/test/nuts/trackingCommands/conflicts.nut.ts
+++ b/test/nuts/trackingCommands/conflicts.nut.ts
@@ -127,7 +127,34 @@ describe('conflict detection and resolution', () => {
   });
 
   it('gets conflict error on push', () => {
-    execCmd<PushResponse>('force:source:push --json', { ensureExitCode: 1 });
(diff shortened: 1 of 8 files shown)


test/nuts/trackingCommands/conflicts.nut.ts
```diff
@@ -127,7 +127,34 @@ describe('conflict detection and resolution', () => {
   });
 
   it('gets conflict error on push', () => {
-    execCmd<PushResponse>('force:source:push --json', { ensureExitCode: 1 });
+    const pushResponse = execCmd<PushResponse>('force:source:push --json', { ensureExitCode: 1 });
```
mshanemc, line 130: nice to have all that detail in here.

mshanemc (approved):

WillieRuemmele (approved):

PR #863, opened 2023-06-08
fix: bump proxy-agent to latest major version

shetzel:
### What does this PR do?
bump `proxy-agent` to latest major version and remove `proxy-from-env` dep since it is now redundant with latest `proxy-agent`

@W-13546885@


mshanemc (approved):
will merge without QA because it's only affecting a deprecated, hidden command

PR #927, opened 2023-08-08
Add GHA prerelease workflows

iowillhoit:
[@W-13907706@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-13907706)


.github/workflows/failureNotifications.yml
```diff
@@ -1,13 +1,12 @@
 name: failureNotifications
-
 on:
   workflow_run:
     workflows:
       - version, tag and github release
```
mshanemc, line 5: why keep this if you deleted that workflow

mshanemc (changes requested):

.github/workflows/failureNotifications.yml
iowillhoit: Cause i forgot 😅

PR #992, opened 2023-11-01
plugin-source ESM

WillieRuemmele:
### What does this PR do?

ESM'ifies plugin-source

@W-14406731@


.lintstagedrc.cjs
cristiand391, line 1: this file can be removed, see: https://github.com/salesforcecli/plugin-login/pull/650#discussion_r1380694430

src/commands/force/mdapi/retrieve.ts
```diff
@@ -100,16 +102,16 @@ export class Retrieve extends SourceCommand {
...
+  protected retrieveResult: RetrieveResult | undefined;
+  private sourceDir: string | undefined;
+  private retrieveTargetDir: string | undefined;
+  private zipFileName: string | undefined;
+  private unzip: boolean | undefined;
+  private wait: Duration | undefined;
+  private isAsync: boolean | undefined;
+  private mdapiRetrieve: MetadataApiRetrieve | undefined;
+  private flags!: Interfaces.InferredFlags<typeof Retrieve.flags>;
+  private org!: Org | undefined;
```
cristiand391, line 114: do you know why this is required now? we already had strict checks enabled before
https://www.typescriptlang.org/tsconfig#strictPropertyInitialization

src/commands/force/mdapi/retrieve.ts
```diff
@@ -224,21 +228,23 @@ export class Retrieve extends SourceCommand {
...
+        waitTime: this.wait?.quantity ?? '',
         verbose: this.flags.verbose ?? false,
         retrieveTargetDir: this.retrieveTargetDir,
         zipFileName: this.zipFileName,
         unzip: this.unzip,
       };
       const formatter = new RetrieveResultFormatter(
         new Ux({ jsonEnabled: this.jsonEnabled() }),
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
cristiand391, line 247: is this still required? if so, can you add a comment about why?

src/commands/force/source/deploy.ts
```diff
@@ -160,9 +163,11 @@ export class Deploy extends DeployCommand {
     junit: Flags.boolean({ summary: messages.getMessage('flags.junit.summary') }),
   };
   protected readonly lifecycleEventNames = ['predeploy', 'postdeploy'];
-  protected tracking: SourceTracking;
+  protected tracking!: SourceTracking;
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
cristiand391, line 168: if this is required to avoid `flags` being possible undefined you could use the non-null assertion:
```diff
-private flags: Interfaces.InferredFlags<typeof Deploy.flags>;
+private flags!: Interfaces.InferredFlags<typeof Deploy.flags>;
````

File: src/commands/force/source/pull.ts [hunk 4]
```diff
@@ -113,7 +115,7 @@ export default class Pull extends SourceCommand {
     if (this.flags['api-version']) {
       componentSet.apiVersion = this.flags['api-version'];
     }
-    const username = this.flags['target-org'].getUsername();
+    const username = this.flags['target-org'].getUsername() as string;
```
cristiand391, line 118: I wonder why `getUsername` says this is optional when the username is always expected: https://github.com/forcedotcom/sfdx-core/blob/b3c1d6b38aed89b063858352a0b9bc845e814a0f/src/org/connection.ts#L77

https://github.com/forcedotcom/sfdx-core/blob/b3c1d6b38aed89b063858352a0b9bc845e814a0f/src/org/org.ts#L844

src/formatters/deployProgressBarFormatter.ts
```diff
@@ -68,7 +73,6 @@ export class DeployProgressBarFormatter extends ProgressFormatter {
 
   // used to initialize the progress bar
   protected initProgressBar(): void {
```
cristiand391, line 75: now that `progressBar` is initialized when defined this can method can be removed.

src/formatters/deployResultFormatter.ts
```diff
@@ -223,6 +227,8 @@ export class DeployResultFormatter extends ResultFormatter {
         this.ux.log('');
         this.ux.styledHeader(chalk.red(`Component Failures [${failures.length}]`));
         this.ux.table(
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
           failures.map((entry: FileResponseFailure) => ({
```
cristiand391, line 232: if you remove the `FileResponseFailure` type it shows almost all of the props it tries to access doesn't exist. Are the types from SDR wrong or where does it get corrected?

src/commands/force/source/pull.ts
mshanemc: The sfdx-core `Org` class allows you to instantiate an Org (the class), assign various props, and then write it for the first time.

I've always wanted to make either an option on Org.create to say, "read this from the FS and throw" and then have better typing on that version.  Or to have a method that verifies it's coming from the FS and then you can depend on all the expected props being there.

It's kind of hard, though, with a bunch of these being define as class props rather than a type-able object.

src/commands/force/mdapi/retrieve.ts
cristiand391: update:
turns out we didn't have strict checks enabled before.

src/commands/force/mdapi/retrieve.ts
cristiand391: fixed

src/formatters/deployProgressBarFormatter.ts
cristiand391: removed

src/formatters/deployResultFormatter.ts
cristiand391: I couldn't get this to work without refactoring a bunch of stuff, will verify on QA.

cristiand391:
**QA notes**:

✅ mdapi/source deploy/retrieve commands, human/json output 
✅ deploy reporter

PR #999, opened 2023-11-07
feat: use new not-config-file stl

mshanemc:
### What does this PR do?
well, see, it all started when I wanted to pull in https://github.com/forcedotcom/source-tracking/pull/496
but then I bumped devScripts, too.
- changes a lot of `node:` 
- requires some changes for enum use
- wanted strictNull checks, which we never managed to get working on this plugin because of its class hierarchy challenges, so that 1 rule is disabled

And hey, we might as well get rid of some oclif.devPlugins and use the `cli:sf` pattern instead
Oh, now we're no longer needing shelljs, so that can go.
Wait, now that we've gotten rid of so many devDeps, it's easy to see some that aren't being used anymore (and probably haven't since the `sfifification` work)

Sorry.

rollout of [@W-14399468@](https://gus.lightning.force.com/a07EE00001dX9wUYAS)


mdonnalley (approved):
This will be a nasty merge with https://github.com/salesforcecli/plugin-source/pull/992

test/nuts/deployDestructive.nut.ts
```diff
@@ -20,6 +20,7 @@ describe('source:deploy --destructive NUTs', () => {
     const pathToClass = path.join(testkit.projectDir, output, `${apexName}.cls`);
     execCmd(`force:apex:class:create --classname ${apexName} --outputdir ${output} --api-version 58.0`, {
       ensureExitCode: 0,
+      cli: 'sf',
```
WillieRuemmele, line 23: if we're using `cli: 'sf'` we can remove PDR as a dev-dep and oclif dev-plugin right?

WillieRuemmele (approved):

PR #1081, opened 2024-02-09
fix: remove an undefined TypeError

mshanemc:
### What does this PR do?
while digging around telemtry because of https://github.com/forcedotcom/cli/issues/2697 I found another TypeError that's happening.

see github comments  for explanation

[skip-validate-pr]

### DIFF
--- a/src/formatters/source/pushResultFormatter.ts
+++ b/src/formatters/source/pushResultFormatter.ts
@@ -14,6 +14,7 @@ import {
   DeployResult,
   FileResponse,
   FileResponseFailure,
+  FileResponseSuccess,
   MetadataResolver,
   SourceComponent,
   VirtualTreeContainer,
@@ -23,7 +24,7 @@ import { ensureArray } from '@salesforce/kit';
 import { Ux } from '@salesforce/sf-plugins-core';
 import { ResultFormatter, ResultFormatterOptions } from '../resultFormatter.js';
 
-Messages.importMessagesDirectoryFromMetaUrl(import.meta.url)
+Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-source', 'push');
 
 export type PushResponse = {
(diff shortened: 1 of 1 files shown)


src/formatters/source/pushResultFormatter.ts
```diff
@@ -132,33 +133,32 @@ export class PushResultFormatter extends ResultFormatter {
     }
     // "content" property of the bundles as a string
     const contentFilePathFromDeployedBundles = this.componentsFromFilenames(
-      // the .filter(isString) should ensure only strings are present, but TS isn't finding that
-      bundlesDeployed.map((fileResponse) => fileResponse.filePath as string)
+      bundlesDeployed.map((fileResponse) => fileResponse.filePath).filter(isString)
```
mshanemc, line 136: SDR has a handler for this now in https://github.com/salesforcecli/plugin-source/pull/1081 but this eliminates the evil asssertion.  `filePath` is typed as `string | undefined` and we'll respect that.

src/formatters/source/pushResultFormatter.ts
```diff
@@ -132,33 +133,32 @@ export class PushResultFormatter extends ResultFormatter {
...
     )
       .map((c) => c.content)
       .filter(isString);
 
     // there may be deletes not represented in the file responses (if bundle type)
     const resolver = new MetadataResolver(undefined, VirtualTreeContainer.fromFilePaths(this.deletes));
-    return (
-      this.deletes
-        .map((filePath) => {
-          const cmp = this.resolveComponentsOrWarn(filePath, resolver)[0];
```
mshanemc, line 146: this [0] would be `undefined` when `resolveComponentsOrWarn` returns an empty array, which it can (and will when the filepath is empty!)

src/formatters/source/pushResultFormatter.ts
```diff
@@ -132,33 +133,32 @@ export class PushResultFormatter extends ResultFormatter {
...
-              filePath,
-            } as FileResponse;
-          }
-        })
-        .filter((fileResponse) => fileResponse)
-        // we can be sure there's no undefined responses because of the filter above
-        .concat(withoutUnchanged) as FileResponse[]
+
+    return withoutUnchanged.concat(
+      this.deletes.flatMap((filePath) =>
```
mshanemc, line 145: by usingFlatMap and doing all the members of the array, we 

1. handle empty array (it'll return `[]` which flatMap eliminates)
2. can be filtered, instead of `if` so that it returns `[]` when there are no bundle types
3. could support the scenario where more than 1 SourceComponent is returned by `resolveComponentsOrWarn`.  Not sure that ever happens, but we were ignoring it.

src/formatters/source/pushResultFormatter.ts
```diff
@@ -132,33 +133,32 @@ export class PushResultFormatter extends ResultFormatter {
...
+              contentFilePathFromDeployedBundles.includes(pathResolve(cmp.content ?? ''))
+          )
+          .map(
+            (cmp) =>
+              ({
+                state: ComponentStatus.Deleted,
+                fullName: cmp.fullName,
+                type: cmp.type.name,
+                filePath,
+              } satisfies FileResponseSuccess)
```
mshanemc, line 159: satisfies lets TS check that it is an ok `FileResponseSuccess` without telling it to `as`sume that it really is if it isn't.

shetzel (approved):

